PR #4, opened 2022-10-05
Download all AWS images from all regions

major:
This provides a one-shot function to download all AWS images from all regions.


FKolwa (approved):
LGTM!

PR #10, opened 2022-10-06
Move fixtures to shared conftest for re-use

major:
These same fixtures will be useful within the update_images tests especially when we split the hourly function into hourly and cloud access.


FKolwa (approved):
LGTM!

Issue #7, opened 2022-10-06
🔧 Refactor AWS hourly images to get cloud access images as well


major:
We have `get_aws_hourly_images()` now, but it pulls a full list of images and filters out the cloud access images. We should probably refactor this to make one function that does the API call and then two separate functions that filter the results. We will need the list of cloud access instances eventually.

PR #11, opened 2022-10-06
Refactor AWS images for hourly/cloudaccess

major:
Update pytest fixtures to include cloud access billing codes and use
those fixtures consistently. Add the billing codes to config and use the
constants consistently.

Split out the actual AWS describe_image call into its own function for
easier mocking and testing. Add an argument to get_aws_images that
specifies which image types to filter for with hourly as the default.

Fixes: #7

### DIFF
--- a/src/rhelocator/update_images.py
+++ b/src/rhelocator/update_images.py
@@ -22,30 +22,50 @@ def get_aws_regions() -> list[str]:
     return sorted([x["RegionName"] for x in raw["Regions"]])
 
 
-def get_aws_hourly_images(region: str) -> list[str]:
+def aws_describe_images(region: str) -> list[str]:
+    """Make an API call to AWS to get a list of RHEL images in a region.
+
+    Args:
+        region: AWS region name, such as us-east-1
+
+    Returns:
+        List of dictionaries containing image data.
+    """
+    ec2 = boto3.client("ec2", region_name=region)
+    raw = ec2.describe_images(Owners=["309956199498"], IncludeDeprecated=False)
+    return raw["Images"]
+
+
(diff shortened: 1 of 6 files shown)


src/rhelocator/update_images.py
```diff
@@ -22,30 +22,50 @@ def get_aws_regions() -> list[str]:
...
-    # RunInstances:0010 == Hourly billing by cloud provider
-    # RunInstances:00g0 == Cloud Access
-    return [x for x in raw["Images"] if x["UsageOperation"] == "RunInstances:0010"]
+    # Filter the results based on the billing code of the image.
+    images = aws_describe_images(region)
+    return [x for x in images if x["UsageOperation"] == billing_code]
 
 
-def get_aws_all_hourly_images() -> dict[str, list[str]]:
+def get_aws_all_images(image_type: str = "hourly") -> dict[str, list[dict[str, str]]]:
```
FKolwa, line 63: Thanks for this. Wanted to implement it myself so this save me loads of time ;)

FKolwa:
I like the changes. About your concern regarding the individual commits: I think the readability is fine but we might want to consider squashing the "fix" commits. e.g. 
78147f2 and b69aa68, a91fafc and c7f7389. 

I don't think splitting the commits is necessary though as they do follow a logical structure that makes sense.

FKolwa (approved):

src/rhelocator/update_images.py
major: Hah yeah, I wasn't thinking through it terribly deeply when I wrote it the first time. 🤦🏻‍♂️

major:
Looks like I have a conflict with that other PR. I'll resolve it right quick.

Issue #12, opened 2022-10-06
Azure functions should get an access_token automatically


major:
Consumers of the Azure image-related functions shouldn't care about the access token. Instead, have the functions call for their own access token.

PR #13, opened 2022-10-06
Make Azure functions call for access token

major:
A consumer of the Azure image/location functions shouldn't need to know about calling the function to get an access_token. Let the Azure functions request an access token as part of their existing work.

Fixes: #12


FKolwa (approved):
I like it. Indeed, this shouldn't be an individual step.

Issue #6, opened 2022-10-06
RHELocator is missing a release workflow


FKolwa:
The RHELocator needs its own release workflow that is not run on every PR / push to main but can be triggered on release (e.g. tag commit merge). The release workflow should use the provided CLI tools to retrieve the image data, format the data and bake the data into a release container image (which can then be pushed to a registry).

FKolwa:
Dibs!

FKolwa:
I like the idea! Looking into it.

PR #15, opened 2022-10-07
Add code release workflow

FKolwa:
This workflow is rather simple and the basis for the container_release workflow.

The code base is build and released under the version tag.
The follow up workflow takes the most recently released version, installs and uses the CLI tools to create the image data json files which are baked into the container image.

Partly fixes: #6


major (approved):
This looks good. Let's give it a try!

PR #20, opened 2022-10-07
remove hardcoded project name

### LINKED ISSUE #19: Remove hardcoded project name from get_google_images
`get_google_images()` in `update_images.py` has a hardcoded project name. That should be moved to `config.py.`

miyunari:
Hey, now `GCP_PROJECTNAME` is hardcoded in `config.py`, should it be an environment variable instead? :)
And do you have any idea, why the CI-pipeline fails? It also fails without a diff^^ :D 

Closes #19 

cc @major @FKolwa

### DIFF
--- a/src/rhelocator/config.py
+++ b/src/rhelocator/config.py
@@ -45,5 +45,11 @@ AZURE_RHEL_IMAGE_TREE = {
     }
 }
 
-# GCP
+#    ________________ 
+#   / ____/ ____/ __ \
+#  / / __/ /   / /_/ /
+# / /_/ / /___/ ____/ 
+# \____/\____/_/ 
+
+# Defines project name.
 GCP_PROJECTNAME = "rhel-cloud"
--- a/src/rhelocator/config.py
+++ b/src/rhelocator/config.py
@@ -44,3 +44,6 @@ AZURE_RHEL_IMAGE_TREE = {
         "RHEL": {"7lvm-gen2": "latest", "8-lvm-gen2": "latest", "9-lvm-gen2": "latest"}
     }
 }
+
+# GCP
+GCP_PROJECTNAME = "rhel-cloud"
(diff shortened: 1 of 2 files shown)


major:
@miyunari Oh, I think this is running from your fork and it's missing the GitHub secrets for AWS auth. I should make a PR to detect that and disable the e2e tests. 😉

major:
@miyunari I made a PR to fix this (hopefully) over in #27

miyunari:
Hey, I rebased onto main, but it still seems to fail :S

major:
Hmm, we might need a better way to determine if the Actions are running from a fork or from the main repo. 🤔

miyunari:
hmm... Do you know why the secrets don't get added when it's a fork or is there a reason for that? :)

FKolwa:
Well in short: Risk of using exploits to obtain tokens / secrets from your Github workflow. Github did fix some common CVEs in the past which allowed users to export values for e.g. the Github token used in your workflow routine but this doesn't stop people from finding creative new ways to misuse triggers like pull_request_target. 

Anyways, interesting thing about the Github context: For the PR from a fork the context seems to be set to the target repo (e.g. redhatcloudx/rhelocator) so checking for repo names or repo owners will always return 'true'.
Also notice the annotation: 'true' apparently the conditional check returns a string value when used inside the if clause of the workflow which will always be evaluated as true as the string isn't empty... I'm proposing a pretty stupid workaround in my next PR until we build a pipeline that executes secret based tests in a different context.

FKolwa:
PR: #30

major:
@FKolwa thanks for #30 -- just merged it.

@miyunari  Want to rebase this one and give it another try?

src/rhelocator/config.py
```diff
@@ -44,3 +44,6 @@
         "RHEL": {"7lvm-gen2": "latest", "8-lvm-gen2": "latest", "9-lvm-gen2": "latest"}
     }
 }
+
+# GCP
```
miyunari, line 48: Hi @major :) 
I noticed you created some ascii art headers for AWS and Azure. How can I do the same with GCP? :D 
I only found this [web page](https://patorjk.com/software/taag/#p=display&f=Blocks&t=GCP), but I am not sure which font I have to use?

major: Hah, I used `figlet` for that. 😉

miyunari: Ah :smile: , thanks! Worked :)

major (approved):
Nice work, @miyunari! 👏🏻

Issue #8, opened 2022-10-06
Azure image retrieval support


major:
Support exists for getting azure's locations (regions), but not images. We need a function for that as well as CLI support.

major:
So it looks like the tree works something like this:

- The lowest level is a `publisher`
- A `publisher` makes `offers`
- Within each `offer` is a set of `skus`
- Each `sku` corresponds to a virtual machine image
- That image can have multiple versions

🥵

major:
Here's what I was able to dig up in eastus: https://gist.github.com/major/7c2c59b875c5126487e512e20de649f1

major:
Let's start with these three SKUs for now:

1. `7lvm-gen2`
2. `8-lvm-gen2`
3. `9-lvm-gen2`

The next question is: are these SKUs the same in each region? Will require some more digging.

major:
Okay, Azure URNs are identical regardless of location/region. I also learned that Azure uses SKU/plan_id interchangeably.

PR #23, opened 2022-10-07
Basic functions for getting images from Azure

major:
This lays the groundwork for getting publishers, offers, skus, and image versions on Azure. It helps with #8 but doesn't fix it quite yet. That will come in a subsequent PR.


FKolwa (approved):
Looks good. We should think about splitting update_images and test_update_images into vendor specific files though.

PR #37, opened 2022-10-11
add gcp image list command to CLI

### LINKED ISSUE #17: CLI: Add Google images support
The CLI must be able to dump RHEL images on GCP.

miyunari:
Hey @major @FKolwa ,

I am a bit confused right now. We assign the registered github secret `secrets.GOOGLE_APPLICATION_CREDENTIALS` to an environment variable `GCP_APP_CREDENTIALS`, but we never use it. 

In the `get_google_images` function we create an `compute_v1.ImagesClient` client, is he supposed to use those credentials? 

I went to the [google docs](https://cloud.google.com/python/docs/reference/compute/latest/google.cloud.compute_v1.services.instances.InstancesClient#google_cloud_compute_v1_services_instances_InstancesClient_aggregated_list), but it also didn't really helped me. :) 

Can you give me some hints? :sweat_smile: 

closes #17

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,8 +88,7 @@ jobs:
           AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
           AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
           AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
-          GCP_APP_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS}}
-
+          GOOGLE_APPLICATION_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS}}
 
       - name: Run pytest with offline tests only
         if: env.SECRET_CHECK == null
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,6 +88,8 @@ jobs:
           AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
           AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
           AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          GCP_APP_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS}}
+
 
       - name: Run pytest with offline tests only
         if: env.SECRET_CHECK == null
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,11 @@ MOCKED_AZURE_IMAGE_VERSION_LIST = [
     "9.0.2022090601",
 ]
 
+MOCKED_GCP_IMAGE_LIST = [
+    {"status": "READY"},
+    {"status": "DEPRICATED"},
+]
+
 
 def pytest_configure(config: any) -> None:
     config.addinivalue_line("markers", "e2e: mark as end-to-end test.")
@@ -52,4 +57,12 @@ def mock_azure_image_versions_latest(mocker):
     """Provide an offline result got get_azure_image_versions."""
     mock = mocker.patch("rhelocator.update_images.get_azure_image_versions")
     mock.return_value = [MOCKED_AZURE_IMAGE_VERSION_LIST[0]]
+    return mock
+
+
+@pytest.fixture
+def mock_gcp_images(mocker):
+    """Provide an offline result for calls to get_google_images."""
+    mock = mocker.patch("rhelocator.update_images.get_google_images")
+    mock.return_value = MOCKED_GCP_IMAGE_LIST
     return mock
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,7 +20,7 @@ MOCKED_AZURE_IMAGE_VERSION_LIST = [
 
 MOCKED_GCP_IMAGE_LIST = [
     {"status": "READY"},
-    {"status": "DEPRICATED"},
+    {"status": "DEPRECATED"},
 ]
 
 
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -87,7 +87,6 @@ def test_aws_regions_offline(mock_aws_regions, runner):
     assert "us-east-1" in parsed
     assert result.exit_code == 0
 
-
 @pytest.mark.e2e
 def test_azure_images_live(runner):
     """Run a live test against the Azure API to get images via CLI."""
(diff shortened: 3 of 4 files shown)


FKolwa:
@miyunari Yup that is correct! It isn't necessary to reference the ENV in code.
In fact you won't find any of the other secrets either (like AWS_ACCESS_KEY_ID).
Most (if not all) cloud provider CLIs authenticate using local configurations that are loaded into ENVs at runtime.
Usually you would use you user credentials to sign into Google Cloud but with an automated service like this a service account can be used. 
The fact that "GOOGLE_APPLICATION_CREDENTIALS" is set for the workflow step means that the gcloud cli will be able to read it and use it to authenticate with the cloud provider API. I'll leave you a link in case you want to read more about Google Application Default Credentials (ADC): https://cloud.google.com/docs/authentication/provide-credentials-adc#local-key

A small hint though: The env needs to be mapped to "GOOGLE_APPLICATION_CREDENTIALS" in the workflow context as well for it to properly work.

Hope this helps a little!

.github/workflows/main.yml
```diff
@@ -84,6 +84,7 @@ jobs:
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          GCP_APP_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS}}
```
FKolwa, line 87: Needs to be mapped to GOOGLE_APPLICATION_CREDENTIALS

FKolwa (changes requested):

miyunari:
Thank you @FKolwa ! That was exactly the information I was looking for :smile: 

But now, there is another issue :sweat_smile:  Unfortunately it's now unclear to me, how to test my changes. 
I got this reference from major, but I don't see the correlation:
https://github.com/redhatcloudx/rhelocator/blob/b97967e77354d0331f0f7bc4a607c00b4b1eea16/tests/test_cli.py#L77-L103

FKolwa:
@miyunari Haha yes this is a bit confusing tbh! 

The gcloud implementation is pretty rough at this point. [get_google_images](https://github.com/redhatcloudx/rhelocator/blob/0e0b13a51330cb213b84f6d4c7ec6399480df5e9/src/rhelocator/update_images.py#L254) currently queries all images in the [rhel-cloud](https://github.com/redhatcloudx/rhelocator/blob/0e0b13a51330cb213b84f6d4c7ec6399480df5e9/src/rhelocator/config.py#L55) project and returns everything that isn't [deprecated](https://github.com/redhatcloudx/rhelocator/blob/0e0b13a51330cb213b84f6d4c7ec6399480df5e9/src/rhelocator/update_images.py#L269).

In the scope of this ticket, my requirement for an end to end test would be 
- query the correct api endpoint by calling get_google_images by using 'runner.invoke'
- parse the json data 
- confirm that all images that are returned do not contain the first level key status with the value 'DEPRECATED'
In this case you can copy / paste most of what @major wrote for the azure test!

For the offline test
- copy the structure of the e2e test you just wrote
- create a new mock for the google images in [conftest](https://github.com/redhatcloudx/rhelocator/blob/0e0b13a51330cb213b84f6d4c7ec6399480df5e9/tests/conftest.py#L35) (you can take a look at the AWS mockups. You need to create a new list of mocked images in a json format and create a new fixture that is passed to you offline test).

Now for the tricky part: How do you know what data structure to expect from the google API?
Well if you call gcloud and query for projects within 'rhel-cloud' you will receive something like this:
`  {
    "architecture": "X86_64",
    "archiveSizeBytes": "4184623872",
    "creationTimestamp": "2022-09-20T16:32:45.492-07:00",
    "description": "Red Hat, Red Hat Enterprise Linux, 9, x86_64 built on 20220920, supports Shielded VM features",
    "diskSizeGb": "20",
    "family": "rhel-9",
    "guestOsFeatures": [
      {
        "type": "UEFI_COMPATIBLE"
      },
      {
        "type": "VIRTIO_SCSI_MULTIQUEUE"
      },
      {
        "type": "SEV_CAPABLE"
      },
      {
        "type": "GVNIC"
      }
    ],
    "id": "2043557223711896434",
    "kind": "compute#image",
    "labelFingerprint": "42WmSpB8rSM=",
    "licenseCodes": [
      "7883559014960410759"
    ],
    "licenses": [
      "https://www.googleapis.com/compute/beta/projects/rhel-cloud/global/licenses/rhel-9-server"
    ],
    "name": "rhel-9-v20220920",
    "rawDisk": {
      "containerType": "TAR",
      "source": ""
    },
    "rolloutOverride": {
      "defaultRolloutTime": "2022-09-25T15:32:42Z",
      "locationRolloutPolicies": {
        "zones/asia-east1-a": "2022-09-22T04:32:42Z",
        ....
        "zones/us-west4-c": "2022-09-25T04:32:42Z"
      }
    },
    "selfLink": "https://www.googleapis.com/compute/beta/projects/rhel-cloud/global/images/rhel-9-v20220920",
    "sourceType": "RAW",
    "status": "READY",
    "storageLocations": [
      "eu",
      "asia",
      "us"
    ]
  }`

At this point we don't extract any specific information from this returned data (like we do for AWS) and this is not within the scope of your ticket so feel free to create a minimal test mockup version of this data structure that only contains something like
`{ "status": "READY" }`

.github/workflows/main.yml
major: Yup, the Google Python SDK will look for that environment variable and read it as a path to the JSON file or as a JSON string itself. 😉

tests/test_cli.py
```diff
@@ -72,3 +72,31 @@ def test_aws_regions_offline(mock_aws_regions, runner):
...
+@pytest.mark.e2e
+def test_gcp_images_live(runner):
+    """Run a live test against the Google Cloud API to get images via CLI."""
+    result = runner.invoke(cli.gcp_images)
+    parsed = json.loads(result.output)
+
+    assert isinstance(parsed, list)
+
+    for image in parsed:
+        expected_keys = ["status"]
```
miyunari, line 85: Does it make sense? It looks so empty :D

miyunari:
> Oh boy, that was my mistake. I must have been looking at two things at the same time and put the wrong variable name in the actions workflow. 🤦🏻‍♂️

@major I think you were right :smile: . I shortened the variable name, because I thought we have to store it in `config.py` and use it somewhere :woman_facepalming:

tests/test_cli.py
FKolwa: That would be a bit restrictive indeed :D 
Probably my fault, should have added more information. 
So you currently test if all your keys are within the expected keys which would fail on live data as the returned data contains more than just the status key. What would make more sense at this point is checking if the keys that are returned contain the key that we want to find in there ("status") with the expected value (not "DEPRECATED").
Something similar to this could work:

`assert {image["UsageOperation"] for image in parsed} != "DEPRECATED"`

As soon as we start extracting the necessary information from the raw data this could be extended to test for all the keys we actually need for the rhelocator. I hope this makes kind of sense!

tests/conftest.py
```diff
@@ -18,6 +18,10 @@
     "9.0.2022090601",
 ]
 
+MOCKED_GCP_IMAGE_LIST = [
+    {"status": "READY"},
```
FKolwa, line 22: Maybe add another line with status = DEPRECATED to see that the data is properly filtered by cli.gcp_images

miyunari: Makes sense - added an entry :)

tests/test_cli.py
miyunari: Yes, made sense to me! Thank you again :)

miyunari:
Oh, I see I have some merge conflicts, will try to resolve

tests/test_cli.py
```diff
@@ -115,3 +114,27 @@ def test_azure_images_offline(mock_azure_image_versions, runner):
...
+
+@pytest.mark.e2e
+def test_gcp_images_live(runner):
+    """Run a live test against the Google Cloud API to get images via CLI."""
+    result = runner.invoke(cli.gcp_images)
+    parsed = json.loads(result.output)
+
+    assert isinstance(parsed, list)
+
+    assert {image["status"] for image in parsed} != "DEPRICATED"
```
FKolwa, line 126: Just a tiny nitpick, I think it's "deprecate" other than that it looks great :)

miyunari: Ah thanks, didn't see that :sweat_smile:

FKolwa (approved):
Thanks for the implementation @miyunari ! Changes look great :+1:

major (approved):
👍🏻

PR #40, opened 2022-10-11
Filter out opt-in AWS regions temporarily

major:
Some regions require an opt-in step and that causes the image listings to fail in those regions. I've requested access to these regions in my account but that request can take a few minutes to a few days.

We can revert this as soon as the opt-in process finishes.


major:
Going to merge this quick fix to help with @FKolwa's branch.

PR #44, opened 2022-10-11
Add e2e AWS test for opt-in region

### LINKED ISSUE #43: CLI end to end tests should query an opt-in AWS region
@FKolwa is working on a branch to dump lots of image data into the container build but the `get_aws_regions` function was returning regions that require an opt-in at AWS. That was unexpected! 😱

Our end-to-end tests should test against at least one opt-in region to ensure this doesn't happen again. 😉

major:
AWS requires an opt-in process for certain regions and this may cause regions to be returned from a `describe_regions()` call that cannot be used along with a `describe_images()` call until the opt-in process is complete.

Fixes #43

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -16,6 +16,20 @@ def runner():
 
 
 @pytest.mark.e2e
+def test_aws_hourly_images_live_opt_in_region(runner):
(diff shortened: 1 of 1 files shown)


FKolwa (approved):

FKolwa:
Thanks for solving this!

PR #45, opened 2022-10-12
Add container release workflow

FKolwa:
This sets the groundwork for a manual release pipeline for building new rhelocator container images.

Things that are missing:
- The actual API endpoint.
- The target we want to push our images to.

Assumptions made:
- The API endpoint will be part of the rhelocator and thus can be installed as part of the rhelocator (poetry installation).
- The API endpoint can be started with a rhelocator cli call.

These items are up for debate. Shoule the API implementation differ from these assumptions, the pipeline can easily be adapted.

### DIFF
--- a/.github/workflows/container_release.yaml
+++ b/.github/workflows/container_release.yaml
@@ -1,0 +1,85 @@
+name: Container-Release
+
+on: [workflow_dispatch]
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    name: Setup rhelocator
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+
+      - name: Bootstrap poetry
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/container_release.yaml
```diff
@@ -0,0 +1,85 @@
...
+          for region in ${regions[@]}; do
+            poetry run rhelocator-updater aws-hourly-images --region $region >> ./results/aws_hourly_images.json
+          done
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+      - name: Build API container
+        id: build-image
+        uses: redhat-actions/buildah-build@v2
```
major, line 57: Oh boy, you get bonus points for using buildah podman! 👏🏻 👏🏻 👏🏻

major (approved):
Great start!

PR #49, opened 2022-10-13
Add Azure image listing to workflow

FKolwa:
Example output: https://github.com/redhatcloudx/rhelocator/actions/runs/3240256203

### DIFF
--- a/.github/workflows/container_release.yaml
+++ b/.github/workflows/container_release.yaml
@@ -40,9 +40,11 @@ jobs:
       - name: Install dependencies
         run: poetry install
 
+      - name: Create output folder
+        run: mkdir -p results
+
       - name: List aws hourly images
         run: |
-          mkdir ./results
           AWS_REGIONS=$(poetry run rhelocator-updater aws-regions)
           regions=$(echo "$AWS_REGIONS" | jq -r '.[]')
           for region in ${regions[@]}; do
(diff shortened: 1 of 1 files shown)


.github/workflows/container_release.yaml
```diff
@@ -40,9 +40,11 @@ jobs:
       - name: Install dependencies
         run: poetry install
 
+      - name: Create output folder
+        run: mkdir ./results >/dev/null 2>&1
```
miyunari, line 44: Why do we need to create the results folder like this upfront? :)

FKolwa: Tbh, just to make sure it's really there. We can create it once we start creating something in that sub-directory, but that would still require us to create the folder before starting to pipe. But honestly there are probably a million more efficient ways to do anything in bash :D

miyunari:
```suggestion
        run: mkdir -p results
```

Would that be the same?

major:
@FKolwa This looks good, but it would be better to use @miyunari's `mkdir -p` suggestion to avoid confusion.

.github/workflows/container_release.yaml
FKolwa: Jap it would ;) Let's do it our way and increase readability!

FKolwa:
> @FKolwa This looks good, but it would be better to use @miyunari's `mkdir -p` suggestion to avoid confusion.

Agreed

Issue #53, opened 2022-10-14
Live test on main branch is broken


miyunari:
Seems like the latest change #37, [broke our CI](https://github.com/redhatcloudx/rhelocator/actions/runs/3249876591/jobs/5332793113) on the main branch.

PR #55, opened 2022-10-14
Dump Google credentials to file

major:
It looks like Google's SDK looks for credentials as a file location only, and not as a JSON string. We need to write it to the filesystem temporarily to use it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,6 +78,11 @@ jobs:
       - name: Install pytest plugin
         run: poetry run pip install pytest-github-actions-annotate-failures
 
+      - name: Dump Google credentials to a temporary file
+        run: echo $GOOGLE_APPLICATION_CREDENTIALS >> /tmp/google-creds.json
+        env:
+          GOOGLE_APPLICATION_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS}}
+
       - name: Run pytest with live e2e tests
         if: env.AWS_ACCESS_KEY_ID != null
         run: poetry run python -m pytest -n auto -p no:sugar -q tests/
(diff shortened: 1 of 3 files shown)


major:
So we probably need to work through the google image listing functions to get more data back from them. I think I just grabbed the `name` as a quick prototype and forgot to go back and get the rest. 🤦🏻‍♂️

FKolwa:
Oh no, yeah it completely slipped my mind that Google expects the service credentials to be located on disk.

Regarding your fix: I feel that service accounts are a bad idea in general and using the application key like this opens a completely new can of worms of security issues (and key management issues, replacing the key regularly, making sure nothing gets dumped ...). 

I've created a PR with a counter proposal that uses Open ID Credentials instead: [PR](https://github.com/redhatcloudx/rhelocator/pull/56)

The PR works using the same service account we already have in place but can easily be modified to use a true "credential less" authentication process using workload identity federations as described here: [Google Blog](https://cloud.google.com/blog/products/identity-security/enabling-keyless-authentication-from-github-actions)

miyunari:
I am not sure if I fully understand both solutions. In this PR we write our non-expiring secret into a file that is used by rhelocator and afterwards removed.
In #56 we relay on a github action that uses github sso to request a new token, that is valid for 1h and used instead? Please corret me if I am wrong :see_no_evil: 

Would both of those solutions fix #53 ?

FKolwa:
@miyunari Yes that is completely correct! The big difference is, as you've already pointed out, the non expiring token that is written to the file system. In this particular case the credentials can't really be misused for anything else than querying image data but it can still authenticate and pretend to be "one of us" so writing it plain text to even an ephemeral container (in which the Github action runs) is still a bit risky. 

The other solution uses our non expiring application key, authenticates on its own with the Google Cloud authentication back end (data is encrypted in transit) and returns that short term token that is written to the file system of the Github action container instead. 
While it looks to be the exact same solution it avoids exposure of our own application key, creates a different user context for the action we execute (it's not the service user that executes the Google API calls, its an instance that uses a token generated by the service user to make the API call) and it self destructs after a certain time frame so we don't have to worry (as much) about leaking it.

Now the interesting part about Open ID Connect (what is used under the hood) is that we can completely avoid using a never expiring service account key and setup an authentication context for our github action instead (now this is probably we can go through in a 1:1. If you want to know more about OpenID I'd recommend the [official page](https://openid.net/connect/) for starters).

Finally to your second question: Yes both solutions solve #53 :D

miyunari:
Thanks @FKolwa ! :smile: Without further knowledge the second solution sounds reasonable.

major:
Closing this as #56 is the better solution.

PR #56, opened 2022-10-17
Add Google Auth to workflow

FKolwa:
This PR removes the need to dump the Google service credentials into a local file before running the e2e tests.
Github enabled [Open ID Connect](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers) last year.

In short:
We can use our credentials as secrets to authenticate with Google, return a one time token (1 hour validity) and use this token to run out API requests.
This PR uses the existing Google service account key to return this token but I will open up a follow up task to move to [Workload Identity Federation](https://cloud.google.com/blog/products/identity-security/enabling-keyless-authentication-from-github-actions) instead.

Github supports OIDC for Azure, AWS and Google as well as external credential providers like HashiCorp's Valut so it might be worth to spend some time and setup a unified authentication system.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,6 +36,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - id: 'auth'
+        name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@v0'
+        with:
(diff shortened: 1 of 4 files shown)


major (approved):
Nice job researching this and implementing it. 👏🏻

PR #60, opened 2022-10-18
Write a credentials file for Google

major:
The docs[0] require a temporary credentials file to be written so that it can be used later with `outputs.credentials_file_path`. The credentials are cleaned up in a post action automatically.

[0] https://github.com/google-github-actions/auth#other-inputs


major:
Nevermind. Writing the credentials file is done by default.

PR #63, opened 2022-10-18
Get more data from GCP about images

### LINKED ISSUE #62: Collect more data from GCP about images
We only collect the image name from GCP right now, but we need some other details, such as the architecture and creation date. The description might be handy, too.

major:
The original proof of concept functions only retrieved the name of the image from GCP but no additional data was included. This patch updates those methods by adding in pertinent information from Google's image list object that could be useful in the locator.

Google returns an unusual object (with more objects under attributes) that is difficult to parse, so it required an additional function to normalize the data and return it similarly to how we handle AWS and Azure.

Fixes #62 

### DIFF
--- a/src/rhelocator/update_images/gcp.py
+++ b/src/rhelocator/update_images/gcp.py
@@ -1,25 +1,58 @@
 """Update images from public cloud APIs."""
 from __future__ import annotations
 
+from typing import Any
+
 from google.cloud import compute_v1
 
 from rhelocator import config
 
-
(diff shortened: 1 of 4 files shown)


FKolwa (approved):
Looking good.

FKolwa:
@major had to manually rebase to fix merge conflicts. I'll go ahead and merge once tests have passed.

major:
Thanks, @FKolwa! 🫂

PR #66, opened 2022-10-20
Rename functions in modules #52

miyunari:
Removed cloud provider name from accessor functions, since its mentioned in the module prefix

closes #52

PR #70, opened 2022-10-21
remove cloud provider name from accessor functions

### LINKED ISSUE #52: Rename functions in cloud provider modules
In #51 we introduced extra modules for aws, gcp and azure. Now function call names are redundant. We should change this.

example:
```
aws.get_aws_regions() => aws.get_regions()
azure.get_azure_images() => azure.get_images()
...
```

miyunari:
same as #66 but from an internal branch

closes #52


major (approved):
Ah, good idea! 👏🏻

PR #72, opened 2022-10-24
Push cloud-image-locator container to quay

FKolwa:
Add secrets and workflow that releases cloud-image-locator container images to quay.
Add GCP cloud image data for completion.

Repo:
https://quay.io/repository/cloudexperience/cloud-image-locator

### DIFF
--- a/.github/workflows/container_release.yaml
+++ b/.github/workflows/container_release.yaml
@@ -3,11 +3,16 @@ name: Container-Release
 on: [workflow_dispatch]
 
 jobs:
-  setup:
+  push:
     runs-on: ubuntu-latest
-    name: Setup rhelocator
+    name: Push Cloud-Image-Locator Container
     steps:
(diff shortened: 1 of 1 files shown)


FKolwa:
pre-commit.ci autofix

miyunari:
Would it make sense to push the image only, when we create a github release? I found [this](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) in the documentation, but I am not 100% sure how it works :sweat_smile:

FKolwa:
Great find @miyunari ;) 
The release trigger does make sense for our code release (which currently only involves tagging). 
The container release on the other hand should happen much more frequently and independent from our code release cycle (e.g. a cron job trigger running once a week).
Though, for now we should stick to manually releasing the container image for our QE team so we won't spam the registry with unusable images.

PR #78, opened 2022-10-24
Allow pre-commit.ci to automatically fix PR

major:
When we submit a commit that has linting issues, [pre-commit.ci](https://pre-commit.ci/) can automatically add a commit to the PR to fix any issues it finds. 🪄

PR #79, opened 2022-10-24
Automatically format docstrings with docformatter

major:
The docformatter project automatically fixes docstrings to match [PEP 257](https://peps.python.org/pep-0257/) requirements.


major:
This should look okay when #78 merges since it has the `autofix_prs` setting set to `true`.

FKolwa (approved):
Thanks for adding this in ;)

PR #81, opened 2022-10-25
pre-commit: Use defaults for isort

major:
Switch back to defaults for the isort tool to more closely align with Python's standards.


FKolwa (approved):
Alright, let's follow the python!

PR #88, opened 2022-10-28
Switch to 3.11 in CI

major:
Let's use the latest released version of 3.11 now that it went GA. 🐍


major:
I'll go ahead and merge this since we were already using 3.11 anyways. 😉

PR #98, opened 2022-11-03
Ignore mypy error for lambda

major:
Also added a docstring for the validation function. 😉

Signed-off-by: Major Hayden <major@redhat.com>

PR #99, opened 2022-11-03
Update dependencies for 2022-11-03

major:


major:
Likely needs #98 to go in first since it has the mypy fix.

PR #100, opened 2022-11-03
Skip Google auth if secrets are not present

major:
When dependabot comes along and tries to bump dependency versions, it lacks access to secrets and the PR always fails. Skip the Google auth step if the Google credentials are unset.


.github/workflows/main.yml
```diff
@@ -36,11 +36,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - id: 'auth'
-        name: 'Authenticate to Google Cloud'
-        uses: 'google-github-actions/auth@v0'
+      - id: "auth"
+        name: "Authenticate to Google Cloud"
+        uses: "google-github-actions/auth@v0"
+        if: env.credentials_json != null
```
FKolwa, line 42: I'm not 100% clear on where the credentials_json comes from in this step. 
Wouldn't a test for env.GOOGLE_APPLICATION_CREDENTIALS suffice?

FKolwa:
Thanks for creating this PR. I will go ahead and close my branch now :D

FKolwa (changes requested):

.github/workflows/main.yml
major: Ah good call. `with` doesn't use env variables, so my approach won't work. 😢

FKolwa (approved):
LGTM!

Issue #83, opened 2022-10-26
Compare cloud provider responses in unit tests 


miyunari:
Currently we are using the pre defined response in our cloud provider unit tests. 

TODO:
- [x] read gcp response from a file 
- [x] read azure response from a file
- [x] read aws response from a file

https://github.com/redhatcloudx/rhelocator/blob/3b24b2e345bc8dc17331f82e723da109c46dd046/tests/update_images/test_azure.py#L30-L47

We want to read a raw http response from a file. That way test responses can be updated independently.

## current architecture
![image](https://user-images.githubusercontent.com/54885993/198022671-6c65ef48-fd18-45ea-986c-ab970741c78a.png)

## proposed architecture
![image](https://user-images.githubusercontent.com/54885993/198022830-4dcd3bde-368b-4a46-a7a1-dfb4015eb686.png)

PR #102, opened 2022-11-04
add azure download script

miyunari:
status: **wip**

part of #83

### DIFF
--- a/src/update_testdata.py
+++ b/src/update_testdata.py
@@ -1,0 +1,33 @@
+# AZURE_CLIENT_ID="" AZURE_CLIENT_SECRET="" AZURE_TENANT_ID="" AZURE_SUBSCRIPTION_ID="" python update_testdata.py
+
+import requests
+import os
+
+from rhelocator import config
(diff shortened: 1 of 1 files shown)


FKolwa:
First things first, thank you for creating this! 
Can't wait to see your first test results.

I've got a very general design question: What is the reason behind recreating the Azure API call (and I assume the AWS and Google calls) instead of using the rhelocator CLI commands? Is this going to be separated from the main locator source in the future and needs to be run individually?

miyunari:
To mock the cloud provider APIs I need the raw http responses, which will be stored in the `testdata` folder. 

As far as I can see the Rhelocator CLI doesn't support storing API responses raw data. We could extend the CLI but I am not sure if it would make sense to add functionality that is only for testing. 

I thought that we maybe shouldn't mix test functionality with business logic. 
What do you think? :)

FKolwa:
Ah now I get where you're going with this! Thanks for clarifying. No I share your opinion of not mixing business and test logic ;)

PR #105, opened 2022-11-07
Convert AWS image data to schema approved format

FKolwa:
- Add new function 'format_image' to convert individual images according to JSON schema.
- Add new function 'format_all_images' to retrieve and convert all images for all regions.

Additional changes:
- Add missing architecture property to JSON schema.

### DIFF
--- a/src/rhelocator/update_images/aws.py
+++ b/src/rhelocator/update_images/aws.py
@@ -114,3 +114,65 @@ def parse_image_name(image_name: str) -> dict[str, str]:
         return matches.groupdict()
 
     return {}
+
+
+def format_all_images() -> object:
+    """Retrieve all AWS images from all regions and return a simplified data
+    representation.
+
+    Returns:
+        JSON like structure containting a list of streamlined image
+        information on all regions.
+    """
+    formatted_images: list[dict[str, str]] = []
+
+    hourly_images = get_all_images()
+    cloudaccess_images = get_all_images("cloudaccess")
+
+    for region in hourly_images:
+        for image in hourly_images[region]:
+            formatted_images.append(format_image(image, region))
+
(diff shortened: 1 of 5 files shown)


src/rhelocator/update_images/aws.py
```diff
@@ -114,3 +114,65 @@ def parse_image_name(image_name: str) -> dict[str, str]:
         return matches.groupdict()
 
     return {}
+
+
+def format_all_images() -> object:
```
FKolwa, line 119: I don't like returning 'object' but the literal type annotation would be 'dict[str, dict[str, list[dict[str, str]]]]' ..
Anybody got a better idea?

major: Haha, no better ideas here. This is where typing gets a little challenging.

src/rhelocator/update_images/aws.py
```diff
@@ -114,3 +114,65 @@ def parse_image_name(image_name: str) -> dict[str, str]:
...
+    """Retrieve all AWS images from all regions and return a simplified data
+    representation.
+
+    Returns:
+        JSON like structure containting a list of streamlined image
+        information on all regions.
+    """
+    formated_images: list[dict[str, str]] = []
```
major, line 127: Not a huge deal, but this should probably be `formatted_images` (two t's)

major (approved):
This looks good to me. It would be nice to get that `formated_images` fixed, but we could put it in a follow-on commit if that's easier to keep things moving along.

src/rhelocator/update_images/aws.py
FKolwa: "The fault, dear Brutus, is not in our <disabled linter>, but in our <ability to write plain English>"

Issue #106, opened 2022-11-07
Add image data formatting to Cloud Image Locator CLI


FKolwa:
Calling the Cloud Image Locator CLI should always output image data according to our defined JSON schema.

After image data conversion has been implemented for AWS, Azure and Google cloud, the added methods need to be added to the CLI commands.

PR #107, opened 2022-11-09
Add aws formatting to cli

FKolwa:
This PR adds image formatting to the aws-hourly-images cli command to output a pre-formatted and validated image file.

part of: #106 

Non issue related fixes necessary for this PR:
- Needed to change schema.validate_json parameter type to arbitrary type of "object"
- Fixed name value in mocked AWS image list for Access image
##########################################################
EDIT:

Okay.. I've been dealing with Azure the entire morning and noticed what a mess it will be to replicate the same structure for Azure. We might need to integrate the image formatting methods into the cloud provider specific "get_image" calls to stay consistent..

### DIFF
--- a/src/rhelocator/cli.py
+++ b/src/rhelocator/cli.py
@@ -9,6 +9,7 @@ from rhelocator import __version__
 from rhelocator.update_images import aws
 from rhelocator.update_images import azure
 from rhelocator.update_images import gcp
+from rhelocator.update_images import schema
 
 
 @click.group()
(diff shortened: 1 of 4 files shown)


major:
> Okay.. I've been dealing with Azure the entire morning and noticed what a mess it will be to replicate the same structure for Azure. We might need to integrate the image formatting methods into the cloud provider specific "get_image" calls to stay consistent..

Seems reasonable. 👍🏻

major (approved):

Issue #94, opened 2022-11-03
Azure: Convert raw image data according to JSON schema


FKolwa:
Implement necessary steps to convert raw Azure image data according to the JSON schema.

PR #115, opened 2022-11-14
Format Azure image data according to schema

FKolwa:
Added:
- parse_image_version method to extract major.minor version and date from Azure image "version" property.
- format_image / format_images to convert raw image data to comply with JSON schema.

Additional changes:
- required "arch" property as part of the azure JSON schema definition.

Solves #94

### DIFF
--- a/src/rhelocator/update_images/azure.py
+++ b/src/rhelocator/update_images/azure.py
@@ -1,6 +1,10 @@
 """Update images from public cloud APIs."""
 from __future__ import annotations
 
+import re
+
+from datetime import datetime
+
 import requests
 
 from rhelocator import config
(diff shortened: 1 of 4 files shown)


src/rhelocator/update_images/azure.py
```diff
@@ -229,3 +231,81 @@ def get_images() -> list[dict[str, str]]:
                     results.append(result)
 
     return results
+
+
+def parse_image_version(image_version: str) -> dict[str, str]:
+    """Parse an AWS image name and return extra data about the image.
+
+    Regex101: https://regex101.com/r/yHB9jJ/1
```
FKolwa, line 239: Please give me a sanity check for this regex. I'm not 100% sure if all RHEL images follow this precise formula but it seems to be consistent within the images we currently retrieve.

major: Haha, nicely done here. It looks like you're leaving off the optional last two digits after the year. Hopefully that's what you intended. 😉

major (approved):
This looks good! I'm glad to see the same function names being re-used here. 👏🏻

src/rhelocator/update_images/azure.py
FKolwa: Haha yeah to be honest I haven't found a real use for the optional digits, but I guess we can revisit this at any time.

PR #123, opened 2022-11-17
test gcp filtering based on real data

miyunari:
This PR does change our gcp unit test. From now it will use real data returned from gcp. Currently this list is generated manually. 

using: 
```bash
gcloud compute images list --format="json" > tests/update_images/testdata/gcp_list_images.json
```

In the future that's what our job should do automatically every night/week. 

part of #83


major (approved):
Thanks for working through this!

PR #127, opened 2022-11-21
AWS read test response from json

miyunari:
part of #83

### DIFF
--- a/tests/update_images/test_aws.py
+++ b/tests/update_images/test_aws.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 from unittest.mock import patch
 
-import pytest
+import pytest, json
 
 from jsonschema import ValidationError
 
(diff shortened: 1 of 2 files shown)


tests/update_images/test_aws.py
```diff
@@ -39,6 +41,24 @@ def test_describe_images() -> None:
...
+    with open("tests/update_images/testdata/aws_list_images.json") as json_file:
+        images = json.load(json_file)["Images"]
+
+    for image in images:
+        data = aws.parse_image_name(image["Name"])
+        assert data["version"]
+        assert data["virt"]
+        assert data["date"]
+        assert data["arch"]
+        assert data["virt"]
```
FKolwa, line 56: Just a duplicate

FKolwa (approved):
Just a small change request to remove the duplicate check for data["virt"], other than that it's looking great!

tests/update_images/test_aws.py
miyunari: oh thank you for realising that one! :smile: Will fix it :+1:

PR #137, opened 2022-11-23
periodically update cloud provider test data

### LINKED ISSUE #84: Create cloud provider http response update job
We would like to automatically update our test data which will be introduced in #83

miyunari:
closes #84 

Example:
https://github.com/redhatcloudx/rhelocator/pull/143

### DIFF
--- a/.github/workflows/update_cloud_data.yaml
+++ b/.github/workflows/update_cloud_data.yaml
@@ -1,0 +1,62 @@
+
+name: Update-Job
+
+on:
+  schedule: # run at midnight
+    - cron: "0 0 * * *"
+
+jobs:
+  update-testdata:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: tests/update_images/testdata
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        # check out all branches
+        fetch-depth: 0
+    - id: "auth"
+      name: "Authenticate to Google Cloud"
+      uses: "google-github-actions/auth@v1"
+      with:
+        credentials_json: "${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}"
+        # If credentials aren't available such as PRs from a fork keep
+        # going with the remainder of the job.
+      continue-on-error: true
+#    - name: 'Az CLI login'
+#      uses: azure/login@v1
+#      with:
+#        client-id: ${{ secrets.AZURE_CLIENT_ID }}
+#        tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+#        subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+#        #secert: ${{ secrets.AZURE_CLIENT_SECRET }}
+    - name: Update cloud provider dumps
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        GOOGLE_APPLICATION_CREDENTIALS: "${{ steps.auth.outputs.credentials_file_path }}"
+      run: |
+       ./update.sh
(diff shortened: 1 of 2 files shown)


tests/update_images/testdata/update.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+az vm image list -f RHEL -l eastus --all > azure_list_images.json
```
FKolwa, line 2: To use the Azure client from your bash script like this you need to run `az login` first.
I'm not sure about the gcloud client but it seems that the last test run failed silently on both and pushed empty json files: https://github.com/redhatcloudx/rhelocator/pull/136/files#diff-ce5b5932601b8d4b43b766097e356bbc1a6da18293462b528f6a8f55fce95f44

FKolwa (changes requested):
Thanks for creating this! :) Just missing a step and some logins, other than that it's looking great!

tests/update_images/testdata/update.sh
FKolwa: K seems like gcloud needs an account setup as well: https://github.com/redhatcloudx/rhelocator/actions/runs/3531503702/jobs/5924749016#step:3:13

FKolwa: Brings up a good point though: From what it seems, the failure of the first two client runs are not propagated to the github actions because the last command (the aws cli) succeeds and the bash script exit code is set to 0. Would be nice if we could safeguard this so that the run fails if any of the cli commands fail.

.github/workflows/update_cloud_data.yaml
```diff
@@ -0,0 +1,49 @@
...
+        fetch-depth: 0
+    - name: Update cloud provider dumps
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+        AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+        AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+        AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+        GOOGLE_APPLICATION_CREDENTIALS: "${{ steps.auth.outputs.credentials_file_path }}"
```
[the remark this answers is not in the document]
miyunari: Aaah - makes sense! Thank you  :D

tests/update_images/testdata/update.sh
miyunari: > Brings up a good point though: From what it seems, the failure of the first two client runs are not propagated to the github actions because the last command (the aws cli) succeeds and the bash script exit code is set to 0. Would be nice if we could safeguard this so that the run fails if any of the cli commands fail.

oh, this message just popped up now! :smile: But that makes sense to me aswell!!

FKolwa: Hm, I would assume that there is a work around for this.. at least for gcloud. I mean we already create the credential JSON file so we should be able to authenticate without another login step.. for Azure however I'm unsure. It would be worth reading up on it before manually login in again. Another option would be to check out if there are official Github Actions for that.

miyunari: > Another option would be to check out if there are official Github Actions for that.

Thanks that was a good hint [[azure-login-action](https://github.com/marketplace/actions/azure-login)]. :smiley: Unfortunately, I have not yet got it to run (azure part) and for now I just commented it out. Once I have changed/written the tests for the azure data, I would try to activate it. Would that be ok? :see_no_evil: Or do you have an idea how to fix it? :/

Google and AWS are now working so far #145 :fireworks:

FKolwa (approved):
Looks good for now. Feel free to merge once you're ready.

PR #150, opened 2022-11-24
Add Azure API endpoint

FKolwa:
- Add Azure route
- Add Azure schema
- Add Azure API tests

Changes unrelated to the issue:
- Had to prefix the gcp blueprint name to avoid conflicts with additional api endpoints.


src/rhelocator/api/schema/azure.yml
```diff
@@ -0,0 +1,64 @@
...
+    examples:
+      images: [
+          {
+            "name": "RHEL 9-lvm-gen2 x64",
+            "arch": "x64",
+            "version": "9.0",
+            "imageId": "redhat:RHEL:9-lvm-gen2:9.0.2022090601",
+            "date": "2022-09-06",
+            "virt": "V2"
+          }
```
miyunari, line 63: Do we need the selflink here aswell? :)

FKolwa: Sadly, no. Azure does not support an image selflink :(

miyunari: Ah ok - perfect :)

Issue #80, opened 2022-10-25
Make image locator more resilient against timeout failure


FKolwa:
Cloud provider API endpoints have to handle a lot of traffic.
We need to make sure our API calls take this into account and establish a sane retry period to make sure we can handle the occasional performance drops.

miyunari:
Hey @FKolwa, 
I tried to write down the current architecture and what we could maybe change to improve the behavior of the ci-pipeline. Would be nice to get some feedback on the document :) 

cc: @major 

1: https://docs.google.com/document/d/1qcBBZPMUFUPhSxtqplYWJt6u9jdPhqwOxVQjaq_N4HE/edit

FKolwa:
Hint: Please add retry (as discussed)

PR #154, opened 2022-11-25
Read azure test response from json

### LINKED ISSUE #83: Compare cloud provider responses in unit tests

miyunari:
This PR does not include testing for the detailed azure image information. I would create an extra issue for this, if thats ok? :)

Closes #83 #80

### DIFF
--- a/tests/update_images/test_azure.py
+++ b/tests/update_images/test_azure.py
@@ -248,14 +248,15 @@ def test_get_all_images(mock_azure_image_versions, mock_azure_image_details):
 
 
 def test_format_image():
-    """Test verifying transformed Azure images into a schema approved format."""
+    """Test verifying transformed Azure images into a schema approved
+    format."""
     images = []
     with open("tests/update_images/testdata/azure_list_images.json") as json_file:
         images = json.load(json_file)
-    
+
     for image in images:
         image["hyperVGeneration"] = "untested"
-        image["version"] += ".2022053014" # untested
+        image["version"] += ".2022053014"  # untested
         data = {"images": {"azure": [azure.format_image(image)]}}
         try:
             schema.validate_json(data)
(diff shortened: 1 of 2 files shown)


tests/update_images/test_azure.py
```diff
@@ -246,23 +248,19 @@ def test_get_all_images(mock_azure_image_versions, mock_azure_image_details):
...
 def test_format_image():
-    """Test transforming a single Azure image into a schema approved format."""
-    mocked_image = {
-        "architecture": "x86_64",
-        "hyperVGeneration": "v2",
-        "offer": "offer",
-        "publisher": "publisher",
-        "sku": "sku",
-        "urn": "publisher:offer:sku:7.6.2020082423",
-        "version": "7.6.2020082423",
```
miyunari, line 257: Any idea why the date is missing in the azure data dump we get from the cli tool? :thinking:

FKolwa: Sorry misread date for data at first (it's friday) :p 
Which CLI tool are referring to?

FKolwa: The mocked_image data above is what the format_image method expects (which itself is a product of two methods 'get_image_versions' and 'get_image_details'). But this raises a good point: I think the format_image methods should be private as it requires image data that is pre-processed. Additionally: Only AWS currently calls format_image everything else uses format_images. I think I'll create a  housekeeping task for this.

FKolwa (approved):
I like the changes and I'm okay with your idea to add the tests in a follow up task.

tests/update_images/test_azure.py
miyunari: Ah great! Thank you for your explanation! :)

PR #156, opened 2022-11-25
Add api server to container

### LINKED ISSUE #131: Implement API server deployment in container release
The source of the API server is already copied into the container we push to quay.
These are the things that are missing:
- production build of the Flask server
- running the server in the container entrypoint

FKolwa:
- aws hourly cli command now outputs images for all regions by default. An individual region can still be supplied by using the --region flag.
- api endpoint execition is added to the container file
- small documentation on building and running the container locally

closes #131

### DIFF
--- a/.github/workflows/container_release.yaml
+++ b/.github/workflows/container_release.yaml
@@ -78,7 +78,7 @@ jobs:
 
       - name: Merge JSON files
         run: |
-          jq -s 'reduce .[] as $item ({}; . * $item)' ./data/* >> image-data.json
+          jq -s 'reduce .[] as $item ({}; . * $item)' ./data/* >> ./data/image-data.json
           rm ./data/temp*.json
 
       - name: Build API container
@@ -107,5 +107,5 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: Upload AWS hourly images
-          path: ./results
+          path: ./data
           retention-days: 7
--- a/.github/workflows/container_release.yaml
+++ b/.github/workflows/container_release.yaml
@@ -46,22 +46,18 @@ jobs:
         run: poetry install
 
       - name: Create output folder
-        run: mkdir -p results
+        run: mkdir -p data
 
       - name: List aws hourly images
         run: |
-          AWS_REGIONS=$(poetry run rhelocator-updater aws-regions)
-          regions=$(echo "$AWS_REGIONS" | jq -r '.[]')
-          for region in ${regions[@]}; do
-            poetry run rhelocator-updater aws-hourly-images --region $region >> ./results/aws_hourly_images.json
-          done
+          poetry run rhelocator-updater aws-hourly-images >> ./data/temp-aws-data.json
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
       - name: List azure images
         run: |
-          poetry run rhelocator-updater azure-images >> ./results/azure_images.json
+          poetry run rhelocator-updater azure-images >> ./data/temp-azure-data.json
         env:
           AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
           AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -62,6 +62,22 @@ http://127.0.0.1:5000/apidocs/ by default.
 The API server will read its configuration on boot from the .env file located in
 `src/rhelocator/api/.env`.
 
+### How to build the API container on your local system
+1. Build the container from within the rhelocator directory:
+```
+buildah bud -f ./Containerfile --format docker --tls-verify=true -t rhelocator:latest .
```
miyunari, line 68: My PC tells me, `buildah` is not installed. :smile: What is the difference to `podman build`?
```suggestion
podman build -f Containerfile -t rhelocator:latest .
```

.github/workflows/container_release.yaml
```diff
@@ -46,22 +46,18 @@ jobs:
...
+        run: mkdir -p data
 
       - name: List aws hourly images
         run: |
-          AWS_REGIONS=$(poetry run rhelocator-updater aws-regions)
-          regions=$(echo "$AWS_REGIONS" | jq -r '.[]')
-          for region in ${regions[@]}; do
-            poetry run rhelocator-updater aws-hourly-images --region $region >> ./results/aws_hourly_images.json
-          done
+          poetry run rhelocator-updater aws-hourly-images >> ./data/temp-aws-data.json
```
miyunari, line 53: That looks way less complicated :smiley:

.github/workflows/container_release.yaml
```diff
@@ -76,10 +72,15 @@ jobs:
...
       - name: List Google Cloud images
         run: |
-          poetry run rhelocator-updater gcp-images >> ./results/gcp_images.json
+          poetry run rhelocator-updater gcp-images >> ./data/temp-gcp-data.json
         env:
           GOOGLE_APPLICATION_CREDENTIALS: '${{ steps.google_auth.outputs.credentials_file_path }}'
 
+      - name: Merge JSON files
+        run: |
+          jq -s 'reduce .[] as $item ({}; . * $item)' ./data/* >> image-data.json
```
miyunari, line 81: oh cool! Once we get detailed azure image information, we can merge them with `jq` too! :smiley:

README.md
FKolwa: Yep both tools are compatible ;) . I think summing up the entire buildah system is a bit too much for a comment but here is the 101: Buildah and podman only differ slightly. Builder only cares about creating OCI (Open Container Initiative) images while podman takes care of the entire lifecycle. Buildah comes with a couple of advantages, like being able to build images that you can run in different container environments without necessarily having something like a docker daemon running on your system. There is plenty more but I think the docs tell it best: https://www.redhat.com/en/topics/containers/what-is-buildah

For your local builds it probably doesn't matter though which you use for building! 
I think I'll add the podman command as well

.github/workflows/container_release.yaml
FKolwa: Yeah was about time, wasn't it? :D

.github/workflows/container_release.yaml
FKolwa: Btw. good thing you commented on this :D ! I've noticed I forgot to add the path to the  pipe output :D

README.md
miyunari: Unfortunately I still don't understand why we can't use podman to build the image?

We use buildah to use the docker format instead of the OCI format[1], right? From what I read, the docker format is the OCI format. 

> Docker's original image format has become the OCI Image Specification, and various open-source build tools support it, including:
>    - BuildKit, an optimized rewrite of Docker's build engine;
>    - Podman, an alternative implementation of Docker's command-line tool;
>    - Buildah, a command-line alternative to writing Dockerfiles;
> Given an OCI image, any container runtime that implements the OCI Runtime Specification can unbundle the image and run its contents in an isolated environment. Docker donated its runtime, runc, to the OCI to serve as the first implementation of the standard.

1: https://www.padok.fr/en/blog/container-docker-oci

FKolwa: Oh, you can use podman just fine (or Docker for that matter) and yes you are completely correct OCI is the default docker format. It's just an example of how you can build the container on your local system for the readme. As Buildah is part of my workflow I used it as an example, that's it ;) . You can use whatever you are comfortable with, but I will delete the buildah line form the readme if that helps!

FKolwa: @miyunari Oh noooooooo I just realized what the confusion was all about. I did not see the `--format` flag in my command (I'd like the blame it on autocomplete but I'm sure it was more a copy-and-paste thing :( ).
Yes you are absolutely right about it making no sense to have that in the docs :) . We want to use OCI and not specify the docker format explicitly (but that's up to who ever wants to build that container). I'll remove that flag ASAP.

About the buildah / podman build thing, I still don't see a harm in having both examples in there but I'm not sad if we remove it. What do you think?

FKolwa: @miyunari I've merged your suggestion, thanks for pointing this out! Guess I need to work on my reading skills.

src/rhelocator/api/server.py
```diff
@@ -2,6 +2,7 @@
 
 from flasgger import Swagger
 from flask import Flask
+from flask_cors import CORS
```
major, line 5: Good idea on CORS. 😉

major (approved):
This looks good! We're getting so close now! 👏🏻

PR #168, opened 2022-11-30
resolve name inconsistency

### LINKED ISSUE #117: Create name consistency with google or gcp images
In our code we often use google or gcp for the google/gcp images. 
We should change that to `google` to be consistent.

miyunari:
closes #117


src/rhelocator/api/routes/google.py
```diff
@@ -15,12 +15,12 @@ def gcp_blueprint(data: list[dict[str, str]]) -> Blueprint:
     Returns:
         Flask blueprint serving the provided data.
     """
-    gcp_endpoint = Blueprint("gcp_api", __name__)
+    google_endpoint = Blueprint("google_api", __name__)
 
-    @gcp_endpoint.route("/gcp", methods=["GET"])
-    @swag_from("../schema/gcp.yml")
+    @google_endpoint.route("/google", methods=["GET"])
```
miyunari, line 20: @FKolwa does it make sense to rename our endpoint too? Wasn't sure about this

FKolwa: Hm. I'm a bit torn on this one too. I'd like to stay as consistent as possible but gcp would probably make more sense as an endpoint.

FKolwa (approved):
LGTM!

major:
Such consistency. 👏🏻

PR #173, opened 2022-11-30
Remove dependabot configuration

major:
Dependabot proposes updates to poetry dependencies in a way that always
fails and renovate can monitor the versions of github actions workflows
for us anyway.


FKolwa (approved):
LGTM!

PR #175, opened 2022-11-30
Make renovate update the lockfile

major:
Renovate's default `rangeStrategy` setting is `replace`. That only updates the pyproject.toml if the new version upstream is outside the range of what pyproject.toml will allow.

By using `update-lockfile` instead, renovate instead updates the lockfile when dependencies change within ranges and then updates the ranges (similar to `replace`) when the ranges change.

https://docs.renovatebot.com/configuration-options/#rangestrategy

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,5 +2,6 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "config:base"
-  ]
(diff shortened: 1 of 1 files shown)


miyunari (approved):
LGTM! :)

major:
Fingers crossed 🤞🏻

PR #190, opened 2022-12-05
Rename google cli call in container release workflow

FKolwa:
Rename the gcp-images to google-images cli command to comply with gcp -> google renaming and fix breaking container release pipeline.


miyunari (approved):
Ah perfect - more for the consistency :smiley_cat: :rocket:

Issue #177, opened 2022-12-01
Extend the amount of image data retrieved from GCP


FKolwa:
Our image filter for GCP images is very restrictive at this moment.
We need to revisit our filtering and retrieve a more complete list of offerings for GCP.

PR #192, opened 2022-12-05
Extend amount of GCP images 

FKolwa:
**What has changed?**
- The amount of Google Cloud Platform images has been extended to include rhel-sap-cloud project offerings.
- Tests have been adapted

**Help needed**
Should the SAP offerings be included in our image data?
Is there anything else that should be included?

Solves (in part) #177


FKolwa:
@major Jup. I think I've klicked "update branch" in github :P .. I wonder if we can deactivate that

PR #214, opened 2022-12-13
Container Release: Replace poetry with release install

### LINKED ISSUE #204: Make locator container run without poetry in production
Currently the entire code base is cloned into our release container.
We should only include our released binaries in the release container and use the build version of the locator to serve the API endpoint.

FKolwa:
Remove poetry dependencies from Containerfile and replace it with release install.
Add rhelocator build stage to release workflow.

Closes #204


major (approved):
👍

PR #219, opened 2022-12-15
Add external product to AWS and Google image names

FKolwa:
Add the external product as a part of the image name.
Adding missing tests to get back to 100% code coverage on AWS and Google update_image methods.

Azure already includes the external product in their image names.


miyunari (approved):
LGTM! :)

PR #268, opened 2023-01-16
register health endpoint

### LINKED ISSUE #203: API server needs health endpoint
*Expected Outcome*
If we want to be able to check the health of our infrastructure properly we need to add a health endpoint to our API that monitors a sane value that ensures the functionality of our API endpoint.
After the tissue is solved, the API should provide a simple endpoint called /health that returns 200 if everything is up and running.

*Hints*
Examples would be:
- connectivity to data storage (DB)
- validity of our in memory data
- reach-ability of our connected back-end (not there yet)

miyunari:
This health check endpoint checks for not existing and empty data. 
If the dataset is corrupt HTTP code 500 is returned. 
If everything is fine HTTP code 200 is returned.

closes #203

### DIFF
--- a/src/rhelocator/api/routes/health.py
+++ b/src/rhelocator/api/routes/health.py
@@ -1,13 +1,11 @@
-from typing import Optional
-
 from flasgger import swag_from
 from flask import Blueprint
-from flask import request
 from flask import Response
 
 
 def health_blueprint(data: dict[str, str]) -> Blueprint:
-    """Creates and returns a Flask blueprint that sanity checks the provided image data.
+    """Creates and returns a Flask blueprint that sanity checks the provided
+    image data.
 
     Args:
         data: Formatted image data dict.
@@ -20,15 +18,16 @@ def health_blueprint(data: dict[str, str]) -> Blueprint:
     @health_endpoint.route("/health", methods=["GET"])
     @swag_from("../schema/health.yml")
     def endpoint() -> Response:
-        """Cloud Image Directory health Endpoint sanity checks the provided image data."""
-        
-        message = "ok" 
+        """Cloud Image Directory health Endpoint sanity checks the provided
+        image data."""
+
+        message = "ok"
         status = 200
         image_data_is_corrupt = False
 
         if data is None:
             image_data_is_corrupt = True
-        
+
         if bool(data) is False:
             image_data_is_corrupt = True
 
@@ -36,8 +35,8 @@ def health_blueprint(data: dict[str, str]) -> Blueprint:
             message = "image data is corrupt"
             status = 500
 
-        return Response("{'message':'"+ message +"'}", status=status, mimetype='application/json')
-
-
+        return Response(
+            "{'message':'" + message + "'}", status=status, mimetype="application/json"
+        )
 
     return health_endpoint
--- a/src/rhelocator/api/routes/health.py
+++ b/src/rhelocator/api/routes/health.py
@@ -29,7 +29,7 @@ def health_blueprint(data: dict[str, str]) -> Blueprint:
         if data is None:
             image_data_is_corrupt = True
         
-        if not bool(data):
+        if bool(data) is False:
             image_data_is_corrupt = True
 
         if image_data_is_corrupt:
--- a/src/rhelocator/api/routes/health.py
+++ b/src/rhelocator/api/routes/health.py
@@ -1,0 +1,41 @@
+from typing import Optional
+
+from flasgger import swag_from
+from flask import Blueprint
+from flask import request
+from flask import Response
+
+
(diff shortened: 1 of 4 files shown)


src/rhelocator/api/server.py
```diff
@@ -47,6 +48,10 @@ def create_app(file_path: str) -> Flask:
         aws_blueprint(image_data["images"]["aws"]), url_prefix="/api"
     )
 
+    app.register_blueprint(
+        health_blueprint(image_data["images"]), url_prefix="/api"
```
miyunari: Currently the api endpoint is reachable under `/api/health` does this make sense? Or should we use just `/health` instead?

src/rhelocator/api/routes/health.py
```diff
@@ -0,0 +1,40 @@
...
+        image_data_is_corrupt = False
+
+        if data == None:
+            image_data_is_corrupt = True
+
+        if image_data_is_corrupt:
+            message = "image data is corrupt"
+            status = 500
+
+        return Response(
```
miyunari, line 36: For now I skipped the schema validation, because I saw that we currently run the json schema validation before creating the blueprints. What do you think?

src/rhelocator/api/schema/health.yml
```diff
@@ -0,0 +1,29 @@
+---
+definitions:
```
miyunari, line 2: Can I replace this with the following?
```yaml
 definitions:
  type: object
  properties:
    message:
      type: string
```

src/rhelocator/api/routes/health.py
FKolwa: /api/health is fine for me.

src/rhelocator/api/schema/health.yml
FKolwa: Yeah that would work. You could also omit the definitions block if you preferre:
```
responses:
  200:
    description:
    schema:
      type: object
        properties:
          message:
            type: string
    examples:
          {
            "message": "ok"
          }
```

src/rhelocator/api/routes/health.py
```diff
@@ -0,0 +1,40 @@
...
+    def endpoint() -> Response:
+        """Cloud Image Directory health Endpoint sanity checks the provided
+        image data."""
+
+        # TODO: Check-(check the value that comes back, ... ?)
+        message = "ok"
+        status = 200
+        image_data_is_corrupt = False
+
+        if data is None:
```
FKolwa, line 29: I like the approach. Maybe add another validity check to the endpoint that runs the JSON schema check on your returned data as well (we can move to a database check once we got that going).

src/rhelocator/api/schema/health.yml
```diff
@@ -0,0 +1,24 @@
...
+  type: object
+  properties:
+    message:
+      type: string
+
+responses:
+  200:
+    description:
+    schema:
+      $ref: '#/definitions/images'
```
FKolwa, line 12: There is no '#/definitions/images' in your yml file anymore! You can simply shorten that ref to '#/definitions' and it will work though. The $ref is just injecting the reference object (in this case definitions/images) at this specific point of the yml structure so you can have less code repitition

tests/api/test_server.py
```diff
@@ -141,3 +141,8 @@ def test_request_azure_with_multi_query(client):
         assert test_query["date"] in image["date"]
         assert test_query["virt"] in image["virt"]
     assert response.status_code == 200
+
+
+def test_request_health(client):
```
FKolwa, line 146: Could you add a failing test case as well, using a client in which no proper image data is provided?

FKolwa (changes requested):

src/rhelocator/api/schema/health.yml
miyunari: oh ups - haha :smile: makes sense, thank you :smiley:

tests/api/test_server.py
```diff
@@ -25,6 +28,29 @@ def client(app):
...
+    empty_app.register_blueprint(
+        health_blueprint(image_data), url_prefix="/api"
+    )
+
+    yield empty_app
+
+
+@pytest.fixture()
+def corrupted_client(empty_app):
+    return empty_app.corrupted_client()
```
miyunari, line 49: Any idea how I can pass the `empty_app` here? I am a bit confused :D

miyunari:
Thanks for your feedback! :)  Now it should work :smile:

tests/api/test_server.py
miyunari: Ah - so close :sweat_smile: thanks for the help! :)

FKolwa (approved):
Good job! Once we got the database figured out we can enhance the health check even more.

PR #275, opened 2023-01-18
Use UBI container in CI

major:
Internal deployments may need this packaged in a UBI container it should work well that way for external deployments, too.


FKolwa:
Hey @major! Just realized that your PR has been opened from your own fork. We adapted the regular PR tests to run from forks but the container release workflow requires access to the cloud credentials. So just FYI: If we want to make changes to our release pipeline we need to open the PR from a rhelocator branch not a fork.

PR #276, opened 2023-01-18
Move mypy to pre-commit

major:
This avoid long delays in GitHub actions while building out the types and it also avoids those situations where your pre-commit results look fine and then mypy explodes in the CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -81,9 +81,6 @@ jobs:
       - name: Install dependencies
         run: poetry install
 
-      - name: Run mypy
-        run: poetry run mypy
-
       - name: Install pytest plugin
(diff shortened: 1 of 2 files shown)


FKolwa:
@major Ahhh.. okay, you're right, mypy seems to ignore what we define in the pyproject.toml.
It seems mypy on its own has issues understanding PEP420 namespaces: https://github.com/pre-commit/mirrors-mypy/issues/5

So as I see it we can either try to add the missing __init__.py (as recommended) or we need to rename our packages. It seems mypy doesn't like the google / azure / aws replication in the api and the upadate_images pacakge.

Any thoughts?

major:
Closing this in favor of #309

Issue #222, opened 2022-12-15
Azure: access tokens for one request are requested multiply times


miyunari:
For every SKU we generate new access tokens. 

Currently [get_image_versions](https://github.com/redhatcloudx/rhelocator/blob/592ccfaa0e19c0694afb11ae49f3a5c4e0f28676/src/rhelocator/update_images/azure.py#L207) and [get_image_details](https://github.com/redhatcloudx/rhelocator/blob/592ccfaa0e19c0694afb11ae49f3a5c4e0f28676/src/rhelocator/update_images/azure.py#L214) internally request an azure access token. 

https://github.com/redhatcloudx/rhelocator/blob/592ccfaa0e19c0694afb11ae49f3a5c4e0f28676/src/rhelocator/update_images/azure.py#L142

https://github.com/redhatcloudx/rhelocator/blob/592ccfaa0e19c0694afb11ae49f3a5c4e0f28676/src/rhelocator/update_images/azure.py#L176

Here we see the loop where both methods get called:

https://github.com/redhatcloudx/rhelocator/blob/592ccfaa0e19c0694afb11ae49f3a5c4e0f28676/src/rhelocator/update_images/azure.py#L199-L215

actions:
- [ ] verify if token can be reused in `get_image_versions` and `get_image_details`
- [ ] verify if we could use one access token to request all SKU information

PR #279, opened 2023-01-19
Reuse Azure access tokens

FKolwa:
**What has been changed?**
- Access_token generation has been moved to get_images method.
- Methods querying Azure endpoints now need to be supplied with a token as parameter.
- Moved post request to its own method.
- Raise SystemExit if request errors are caught that would otherwise limit the CLI functionality.

If your are okay with the modifications, the other API calls will be modified as well (in a follow up task).

Closes: #222

### DIFF
--- a/src/rhelocator/update_images/azure.py
+++ b/src/rhelocator/update_images/azure.py
@@ -7,16 +7,18 @@ import time
 from datetime import datetime
 
 import requests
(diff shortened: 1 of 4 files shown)


major (approved):
Nice! This looks like it's going to be much more reliable.

PR #300, opened 2023-01-30
verify if skus are unique per offer

### LINKED ISSUE #221: Test: check that azure images of offers have unique SKU's
In #191 we removed our azure image sanity check. 

Previously we tested if SKU's are unique. This test failed when we introduced the support for multiply offers. Since SKU's are just unique for one specific offer. 

We should create a test to verify that SKU's of specific offers are unique.

https://github.com/redhatcloudx/rhelocator/pull/191/files#diff-cc50f1e74f7e564801dc385ef8a5a27563f8d4d892e319d4feb1bf3e2ae4e50dL233-R234

miyunari:
closes #221

### DIFF
--- a/tests/update_images/test_azure.py
+++ b/tests/update_images/test_azure.py
@@ -260,6 +260,20 @@ def test_get_latest_images(mock_azure_image_versions_latest, mock_azure_image_de
         assert image["urn"] == ":".join(urn_sections)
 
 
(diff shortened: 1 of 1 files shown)


FKolwa (commented):
Looking good! That'll absolutely increase trust in what we deliver :+1:

PR #309, opened 2023-02-01
Move mypy to pre-commit

major:
This avoid long delays in GitHub actions while building out the types and it also avoids those situations where your pre-commit results look fine and then mypy explodes in the CI.

Clean up one "any" type declaration in conftest.py that isn't needed any longer.

Removed several requirements from pyproject.toml since only pre-commit needs them during its run.


FKolwa (approved):
Sweet, looking good! This will certainly keep us from forgetting about poor mypy all the time ;)

PR #318, opened 2023-02-02
poetry: remove pylint

major:
There's no need for pylint in poetry's requirements since it is only used inside the pre-commit virtual environments.


FKolwa (approved):
Good thinking! :+1:

PR #328, opened 2023-02-07
Enable automerge for renovate updates

major:
If lock file updates are needed, renovate will propose a PR and instruct github to automatically merge the PR when tests pass. Also, any minor package updates will be auto-merged unless they change a major version.

Change the update strategy to "bump" so we can be more specific about the versions we use.


FKolwa (approved):
Great idea! That'll save some time.

PR #329, opened 2023-02-07
Remove yesqa/flake8

major:
We keep having issues getting flake8 up to version 6 and it appears that many of the extra flake8 modules used by yesqa have conflicts with it. Also, some of those modules aren't maintained frequently.

Let's remove both and rely on black/isort/mypy/pycln to keep us linted.


miyunari:
Makes sense! :smile:

miyunari (approved):

PR #341, opened 2023-02-10
fix azure tests

### LINKED ISSUE #307: Bug: Offline tests executed from forks fail Azure authentication step
Example: https://github.com/redhatcloudx/rhelocator/actions/runs/4058595003/jobs/6992364595

Check the authentication mock and make sure that the offline tests complete without throwing an "Unable to authenticate" exception.

- [x] fix azure mocks
- [x] fix aws mocks
- [x] fix google mocks
- [x] remove live tests

miyunari:
This PR should fix the azure tests so that authentication is no longer required for these tests.

part of #307

### DIFF
--- a/src/cloudimagedirectory/update_images/azure.py
+++ b/src/cloudimagedirectory/update_images/azure.py
@@ -8,6 +8,8 @@ from datetime import datetime
 
 import requests
 
+import unittest
+
 from requests import HTTPError
 from requests import RequestException
 from requests import TooManyRedirects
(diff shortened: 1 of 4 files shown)


FKolwa (approved):
Nice work, looks good to me!

miyunari:
Yes, we can do that - but then I think it would make sense to do it for all the tests :) 
So all the Mock data into files with scripts to update the data automatically

PR #348, opened 2023-02-13
remove live tests

miyunari:
In this PR, the live tests for the cloud providers were removed to allow it to work with forks.

part of #307


FKolwa (approved):
LGTM!

PR #355, opened 2023-02-15
Enable logging in CLI tool

### LINKED ISSUE #226: Implement logging procedure in cli
*Expected Outcome*
After implementing this new feature, the cli should be able to log info / debug level information depending on its current execution mode.

FKolwa:
This PR improves logging for our CLI tool. This is a jumping off point to refine as we go.
I've added an implementation example to azure.py.

Rich has been added as a dev dependency as it is "too chatty" when it comes to stack traces.
We can consider to improve on this with custom log message formatting if we choose to do so (e.g. scan event_dict for secrets and remove them from the param list).

Errors are re-raised after logging them in the CLI endpoint to maintain exit codes. Stack trace is preserved. Probably worth discussing though.

Closes #226

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -643,14 +643,14 @@ protobuf = ">=4.21.6"
 
 [[package]]
 name = "identify"
-version = "2.5.17"
+version = "2.5.18"
 description = "File identification library for Python"
 category = "dev"
 optional = false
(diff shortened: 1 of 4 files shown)


major (approved):
This looks super helpful! :clap:

PR #412, opened 2023-03-20
add workflow for transformer

### LINKED ISSUE #404: Add workflow file for transformer
We should add a ci-workflow for the transformer to continuously update our image data in the s3 bucket

miyunari:
closes #404


.github/workflows/transformer.yaml
```diff
@@ -0,0 +1,30 @@
...
+  parser:
+    name: "Transform image data"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: "actions/checkout@v3"
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: arn:aws:iam::426579533370:role/github_actions_image_retriever
```
miyunari, line 22: Can we use this role here too? Or do we need to create another one [here](https://github.com/redhatcloudx/infrastructure/blob/ad20f2d955eb621911a9acfafada4cb805f5d782/retriever-poc-iam.tf)?

FKolwa: You need to reference the transformer role: https://github.com/redhatcloudx/infrastructure/blob/ad20f2d955eb621911a9acfafada4cb805f5d782/transformer-poc-iam.tf

It's basically identically with the small difference that the github user content is set to the directory repository:
`   condition {
      test     = "ForAnyValue:StringLike"
      variable = "token.actions.githubusercontent.com:sub"
      values   = ["repo:redhatcloudx/cloud-image-directory:*"]
    }`

so just switch the role name to `github_actions_image_transformer` and you should be good to go!

miyunari: Nice, thank you! :)

FKolwa (approved):
Looking good! Let's merge and give it a try!

PR #416, opened 2023-03-22
Upload JSON data with correct mime type

major:
Allow browsers to handle JSON properly and display it rather than using a binary mime type which requires a download.


miyunari (approved):
Nice! :smiley:

PR #419, opened 2023-03-22
support transforming google images

### LINKED ISSUE #405: Support for transforming google images
Implement the `TransformerGCP` function to `transform.py`, where the google images will be parsed into our schema.

miyunari:
closes #405

### DIFF
--- a/src/cloudimagedirectory/transformer.py
+++ b/src/cloudimagedirectory/transformer.py
@@ -56,5 +56,5 @@ def run(origin_path: str, destination_path: str, api: str, arg_files: str) -> No
         version_prefix = api + "/"
     for result in results:
         result.filename = destination_path + "/" + version_prefix + result.filename
-        #print("put content to filesystem - filename: " + result.filename)
+        # print("put content to filesystem - filename: " + result.filename)
         origin_connection.put_content(result)
--- a/src/cloudimagedirectory/transformer.py
+++ b/src/cloudimagedirectory/transformer.py
@@ -56,5 +56,5 @@ def run(origin_path: str, destination_path: str, api: str, arg_files: str) -> No
         version_prefix = api + "/"
     for result in results:
         result.filename = destination_path + "/" + version_prefix + result.filename
-        print("put content to filesystem - filename: " + result.filename)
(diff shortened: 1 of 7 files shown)


src/cloudimagedirectory/transformer.py
```diff
@@ -56,5 +56,5 @@ def run(origin_path: str, destination_path: str, api: str, arg_files: str) -> No
         version_prefix = api + "/"
     for result in results:
         result.filename = destination_path + "/" + version_prefix + result.filename
-        print("put content to filesystem - filename: " + result.filename)
+        # print("put content to filesystem - filename: " + result.filename)
```
major, line 59: Can we remove this line? :wink:

major: Otherwise this looks good.

tests/transformer/google_test.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+
+def test_transformer_google():
```
FKolwa, line 4: Could you add a docstring to your tests as well as breaking test cases?

miyunari: Yes, I can add a docstring. How would you model a breaking test? The only thing that I have in mind is, that the transformer would fail if we provide an invalid input file. Does this make sense to you?

src/cloudimagedirectory/transformer.py
miyunari: Sure, I will remove it :smile:

tests/transformer/google_test.py
FKolwa: Yeah that makes total sense to me. I would add
- Files with a broken structure (test case: the retriever created a malformed file)
- Files cannot be downloaded (test case: S3 bucket is down / cannot be reached)
I want to make sure that we can catch those failures should they occur.

FKolwa (approved):

tests/transformer/google_test.py
FKolwa: True. Strike the download! We'll cover empty inputs with the rest of the tests in the follow up task!

PR #433, opened 2023-03-27
transfomer: ignore azure images with invalid timestamp

miyunari:
Belongs to https://github.com/redhatcloudx/cloud-image-directory/issues/403

See: https://github.com/redhatcloudx/cloud-image-directory/actions/runs/4530591649/jobs/7979718756#step:11:32

### DIFF
--- a/src/cloudimagedirectory/transform/transform.py
+++ b/src/cloudimagedirectory/transform/transform.py
@@ -166,12 +166,16 @@ class TransformerAZURE(Transformer):
 
                 content["hyperVGeneration"] = "unknown"
 
-                image_data = azure.format_image(content)
-                image_name = image_data["name"].replace(" ", "_").lower()
-                data_entry = connection.DataEntry(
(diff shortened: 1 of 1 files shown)


src/cloudimagedirectory/transform/transform.py
```diff
@@ -162,20 +162,25 @@ def run(self, data):
...
 
-                results.append(data_entry)
+                    results.append(data_entry)
+                except:
+                    print(
+                        "Could not format image, sku: "
+                        + content["sku"]
+                        + " offer: "
+                        + content["offer"]
+                    )
```
major, line 184: Good idea here. :wink:

FKolwa: Agree, good thinking to log that out!

FKolwa (approved):
Useful addition! We might want to add logging to the transformer cli as we did with the updater so we can increase traceability.

PR #439, opened 2023-03-28
transformer: add test for generating list sorted by date

miyunari:
This PR still doesn't solve the weird listing by date in the s3 bucket. But at least it verifies that the `TransformerIdxListImageLatest` works as expected :)

### DIFF
--- a/tests/transformer/test_transform.py
+++ b/tests/transformer/test_transform.py
@@ -16,14 +16,16 @@ def test_transformeridxlistimagelatest():
             },
         ),
         connection.DataEntry(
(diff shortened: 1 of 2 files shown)


tests/transformer/test_transform.py
```diff
@@ -0,0 +1,91 @@
+from cloudimagedirectory.connection import connection
+from cloudimagedirectory.transform import transform
+
+
+def test_transformeridxlistimagelatest():
```
FKolwa, line 5: I :heart: tests

FKolwa (approved):

PR #444, opened 2023-03-29
Transformer add region

miyunari:
I added a region field to all the lists. Now I am a bit unsure if we still need the `ref` entry? Since `provider` + `region` + `name` should be equal to `ref`. :thinking:

cc @FKolwa


miyunari:
Thank you :hugs:  
Yes, I think optimizing this at some point would make sense and leave it for now :relaxed:

PR #446, opened 2023-03-29
transformer: add index page file

miyunari:
With the new file we should have this path: `v1/idx/list/sort-by-date/pages`  with this content:
```
{
  "first": 0,
  "last": 360,
  "entries": 50,
}
```

This file should help us to know for the frontend how many pages we want to serve.


FKolwa (approved):
Nice! The index for the index!

PR #461, opened 2023-04-04
Transformer add date filter

miyunari:


src/cloudimagedirectory/transformer.py
```diff
@@ -41,17 +54,27 @@ def run(origin_path: str, destination_path: str, api: str, arg_files: str) -> No
...
         print("input: " + file.filename)
 
+    filters = [
+        filter.FilterImageByFilename("test"),
+        filter.FilterImageByFilename("beta"),
+    ]
+
+    if filter_until == "default":
+        # NOTE: Subtract 2 years from current time
+        filter_after = datetime.datetime.now() - datetime.timedelta(days=2 * 365)
```
FKolwa, line 64: readability ++

FKolwa (approved):
Looking good! Let's get rid of those old images

PR #466, opened 2023-04-04
transformer: add idx list with image names

### LINKED ISSUE #465: transformer: add idx list with image names
We should create an index list containing all the image names so that we can integrate a search function in the frontend.

miyunari:
closes #465


FKolwa (approved):
Nice! Is this already being pushed with the existing workflow?

PR #541, opened 2023-05-16
cleanup: remove old server api and cli

### LINKED ISSUE #481: Cleanup repository
This repository contains a lot of legacy code that isn't needed anymore.
We should use the time before our next milestone to clean house.

What can be removed?
/src/cloudimagedirectory/...
- API (We have moved away from our back-end server for now. The entire implementation can be cleaned.)
- update_images/aws/azure/google.py (This one is a bit tricky. We need to take a closer look at what functionality we still need and extract those functions into separate helper classes that can be called from within the transformer. schema.py is still in use.)
- cli.py (We should consider either removing this file or merging with the transformer.py)

blocked by #483

miyunari:
In this PR I removed everything I think we don't need anymore. The functionality we still use from `update-images` is already in a separate class https://github.com/redhatcloudx/cloud-image-directory/pull/484. 

closes #481


FKolwa:
Looks good to me. I haven't seen anything worth keeping. 
From my side, we can ignore the pre-commit check this time as this PR is not introducing anything new.

FKolwa (approved):

FKolwa:
Maybe another idea: How about we rename the repo to transformer or image-transformer? The old name doesn't really fit anymore.

miyunari:
> Maybe another idea: How about we rename the repo to transformer or image-transformer? The old name doesn't really fit anymore.

Yes, I have no strong opinion to this :smile: Let's discuss it in the next meeting :+1:

Issue #543, opened 2023-05-17
Add api documentation


miyunari:
We currently provide an api, but it isn't documented. We should have a documentation to help others understand or use our project better.

miyunari:
Swagger ui is now available here: https://imagedirectory.cloud/swagger/index.html

PR #548, opened 2023-05-19
document api with swagger

miyunari:
This PR provides the swagger definition to generate api docs.

**Example**
image

Since swagger-ui consists out of javascript + css does it make sense to ship it via our CDN?

Here is an overview of the needed files:
https://github.com/miyunari/vacadm/tree/4fba6f692c073daf60e5d5097ecce1e7f916b38e/assets/swagger

Part of #543

### DIFF
--- a/docs/api/swagger.yaml
+++ b/docs/api/swagger.yaml
@@ -1,0 +1,290 @@
+openapi: 3.0.3
+info: 
+  title: "Cloud Image Directory"
+  description: |-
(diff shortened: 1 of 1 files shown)


major (approved):
Excellent idea. Let's get this in and keep iterating on it. 👏

FKolwa:
Sweet, looking real good! This will make discussion our API much easier.

PR #558, opened 2023-05-23
e2e test adjustment

### LINKED ISSUE #441: Cleanup transformer e2e tests
Adapt the transformer e2e tests to this style https://github.com/redhatcloudx/cloud-image-directory/pull/427

miyunari:
In this PR I have worked on our e2e tests to make them more Python style. I left out `idx_image_latest_test.py` because I think we will be changing some things there in the next few days and I want to adapt this one after the changes.

closes #441

### DIFF
--- a/tests/transformer/aws_test.py
+++ b/tests/transformer/aws_test.py
@@ -1,6 +1,6 @@
 """Tests for the AWS transformer."""
-import os
 import filecmp
+import os
 
 from cloudimagedirectory import transformer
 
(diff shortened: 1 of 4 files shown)


tests/transformer/aws_test.py
```diff
@@ -1,15 +1,33 @@
...
-        " ${PWD}/tests/transformer/testdata/output/aws/af-south-1/rhel_6.10_hvm_x86_64_hourly2"
+
+    # Ensure the directory was made.
+    assert os.path.isdir(f"{tmp_path}/output/aws/af-south-1")
+
+    # Get current directory
+    pwd = os.getcwd()
+
+    # Check image data by comparing the expected file and the output file byte by byte.
+    filecmp.cmp(
```
FKolwa, line 30: Maybe a stupid question (might be handled by pytest magic I don't know about) but without asserting the returnvalue of filecmp, do these tests even fail on "false"?

miyunari: It's not a stupid question, I was asking myself the same at some point :) I tried it out and read through the documentation of filecmp :) 

In the [documentation ](https://docs.python.org/3/library/filecmp.html) they say that `filecmp.cmp` compares the two files and if they are equal they return true per default, if not they return false. For some reason then the test also failes, but you are right - I wonder why the return value is not ignored. Maybe I will make it more explicit by adding something like this: 
```suggestion
    assert filecmp.cmp(
```
What do you think?

FKolwa: Nice thanks for the information! Yeah adding the assert Infront wouldn't hurt and the rest would be a bit explicit. I like that.

FKolwa (approved):
LGTM! :+1:

PR #566, opened 2023-05-30
fix google selflink

### LINKED ISSUE #547: Google Cloud selflink contains hardcoded project name
This bug was reported by @poncovka :

**Current behavior**
The selflink contained within Google Cloud images only works for images within the "rhel-cloud" project.
Once you try to access images within other projects the reference will fail with an "image not found" response.

Example:
Example image data:
https://imagedirectory.cloud/images/v1/google/global/rhel_8.2_sap_x86_64

Selflink contained in image data:
https://console.cloud.google.com/compute/imagesDetail/projects/rhel-cloud/global/images/rhel-8-2-sap-v20230509

Fixed selflink:
https://console.cloud.google.com/compute/imagesDetail/projects/rhel-sap-cloud/global/images/rhel-8-2-sap-v20230509

**Proposed fix**
The project name is currently hard coded within our transformer code:
https://github.com/redhatcloudx/cloud-image-directory/blob/main/src/cloudimagedirectory/format/format_google.py#L59
This should be turned into a variable value based on the image data provided.

miyunari:
In this PR I removed the hardcoded project name within the Google selflink. I was able to extract the Google project name from the original selflink within the raw image data.

closes #547

### DIFF
--- a/src/cloudimagedirectory/format/format_google.py
+++ b/src/cloudimagedirectory/format/format_google.py
@@ -40,6 +40,7 @@ def image_rhel(image: dict[str, str]) -> dict[str, str]:
     arch = image["architecture"]
     image_id = image["name"]
     date = image["creation_timestamp"]
+    selflink = image["selfLink"]
     extprod = additional_information["extprod"]
     version = additional_information["version"].replace("-", ".")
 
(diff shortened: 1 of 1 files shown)


major (approved):

src/cloudimagedirectory/format/format_google.py
```diff
@@ -55,8 +56,18 @@ def image_rhel(image: dict[str, str]) -> dict[str, str]:
...
+    if len(selflink_list) < 7 or selflink_list[5] != "projects":
+        # NOTE: We extract the current project name from the original selflink.
+        # Example selflinks:
+        # https://.../compute/imagesDetail/projects/rhel-cloud/.../...
+        # https://.../compute/imagesDetail/projects/rhel-sap-cloud/.../...
+        raise Exception("invalid selflink")
+
+    project_name = selflink.split("/")[6]
+
     selflink = "https://console.cloud.google.com/compute/imagesDetail/"
```
poncovka, line 69: Out of curiosity, why don't you just keep the original selflink? It looks like the front-end is using your selflink for the "Launch now" button, but then [recreates the original selflink](https://github.com/redhatcloudx/cloud-image-directory-frontend/blob/26177ac85e27026ce0b80bfb0e13257309ff0bc1/src/app/components/DetailsView.tsx#L102-L108
) for the "Launch from CLI" section. That doesn't look optimal.

Also, the selflink is really important for the actual identification of the image. The image id is not a reliable identifier, so I don't recommend to mess with this attribute.

PR #574, opened 2023-06-02
Hotfix to enable ZST encoded archives loaded by the transformer

FKolwa:
The retriever has been switched to provide zstandard encoded raw data instead of pure JSON.
As this is indeed an archived format it cannot be read like pure JSON data.
To enable the transformer to work with the new data we need to add the proper file conversion.

These changes are up for debate. However we need to fix this within the day as the scheduled transformer run for tomorrow morning will break the data again otherwise.

I've also removed the delete statement in our S3 upload as this removes files that aren't within the upload folder but are present in the S3 bucket.


.github/workflows/transformer.yaml
```diff
@@ -62,5 +62,5 @@ jobs:
 
       - name: Upload to S3
         run: |
-          s3cmd sync --acl-public --delete-removed --mime-type application/json \
+          s3cmd sync --acl-public --mime-type application/json \
```
miyunari, line 65: I see why you think this could be a good idea. But we actually need to remove the files that are not uploaded, otherwise we can't filter things in the transformer, since they will still be on S3 :) WDYT?

FKolwa: Ah I see, I had no idea. Yes that makes sense to me.

Issue #492, opened 2023-04-21
Incorrect metadata of images with the same generated name


poncovka:
CID doesn't use unique references to images, so some of the data get lost during the transformation of the raw data. For example, the https://imagedirectory.cloud/images/v1/aws/eu-west-2/rhel_9.0.0_hvm_x86_64_hourly2 reference seems to be used by four different images called:

* RHEL-9.0.0_HVM-20230313-x86_64-43-Hourly2-GP2
* RHEL-9.0.0_HVM-20221027-x86_64-1-Hourly2-GP2
* RHEL-9.0.0_HVM-20230127-x86_64-24-Hourly2-GP2
* RHEL-9.0.0_HVM-20220513-x86_64-0-Hourly2-GP2
 
As a result, the website sometimes shows incorrect data about images. For example, these are two different entries that show metadata about the same image, however dates suggest that these are two different images.

![Screenshot from 2023-04-20 18-53-14](https://user-images.githubusercontent.com/7396092/233604039-7f96a27f-4d27-42a2-9b24-846551b50caa.png)

![Screenshot from 2023-04-20 18-51-09](https://user-images.githubusercontent.com/7396092/233604062-cbb733b1-bce9-44f1-9236-a26d4e960f64.png)

References to images are names generated by CID based on the image metadata.

https://github.com/redhatcloudx/cloud-image-directory/blob/89403e384471803990dc57961a2188713b9b32a5/src/cloudimagedirectory/format/format_aws.py#L64

PR #576, opened 2023-06-02
transformer: Provide only the latest images

poncovka:
If there are multiple images of the same name, use only the latest one. This is a workaround of an issue with the CID API, that generates files for these images at the same path and looses data about images as a result. Let's throw away the data anyway, but this time intentionally.

Resolves: #492

### DIFF
--- a/src/cloudimagedirectory/filter/filter.py
+++ b/src/cloudimagedirectory/filter/filter.py
@@ -4,6 +4,14 @@ import pandas as pd
 import pytz  # type: ignore
 
 
+def get_utc_datetime(date_string):
+    """Get a timezone-aware comparable UTC datetime object.
+
+    Returns: A datetime object representing the date string.
+    """
+    return pd.Timestamp(date_string).replace(tzinfo=pytz.UTC)
+
+
 def FilterImageByFilename(word: str) -> Callable:
     """Filter images by filename."""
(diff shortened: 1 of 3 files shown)


src/cloudimagedirectory/filter/filter.py
```diff
@@ -20,6 +28,44 @@ def FilterImageByLatestUpdate(latestDate: pd.Timestamp) -> Callable:
...
         for d in data
-        if d.content is not None
-        and pd.Timestamp(d.content["date"]).replace(tzinfo=pytz.UTC) > latestDate
+        if d.content is not None and get_utc_datetime(d.content["date"]) > latestDate
     ]
+
+
+def FilterImageByUniqueName() -> Callable:
+    """Filter latest images with unique names."""
+    print(f"filter images by unique names")
```
miyunari, line 37: Is this a debug leftover? :)

miyunari (approved):
LGTM! :D

FKolwa (approved):
Looks good indeed. Thanks for solving this issue!

src/cloudimagedirectory/filter/filter.py
poncovka: I've removed `f`, because the string doesn't have to be an f-string. Otherwise, all filters have a `print` statement.

PR #581, opened 2023-06-05
fix keep original google selflink

### LINKED ISSUE #588: Make google image data consistent with other cloud providers
Current situation:
{
"name": "RHEL 7 X86_64", 
"arch": "X86_64",
"version": "7", 
"imageId": "rhel-7-v20230509", 
"date": "2023-05-09T15:09:02.196-07:00", "
selflink": "https://console.cloud.google.com/compute/imagesDetail/projects/rhel-cloud/global/images/rhel-7-v20230509"
}

- imageId: The value in this field is not enough to run a GCP image.
- selflink: Is currently parsed from the Google API URL and once again used to create the Google API URL in front-end ( converted twice as discovered by @poncovka and fixed by @miyunari in #581 )

To comply with #581 these changes need to be introduced to the front-end: 
https://github.com/redhatcloudx/cloud-image-directory-frontend/compare/main...google-selflink-switch-%23102

These changes come with a few caveats:
- We need to introduce new conditionals just for google. 
- It implies that the meaning of the "selflink" property won't be consistent across cloud providers.
- It does not solve the issue that our imageId property does not contain the necessary information.

Proposed changes:
{
"name": "RHEL 7 X86_64", 
"arch": "X86_64",
"version": "7", 
**"imageId": "https://www.googleapis.com/compute/v1/projects/rhel-cloud/global/images/rhel-7-v20230509",** 
"date": "2023-05-09T15:09:02.196-07:00", "
selflink": "https://console.cloud.google.com/compute/imagesDetail/projects/rhel-cloud/global/images/rhel-7-v20230509"
}

The imageId should contain the googleapis.com URL to stay consistent with the data provided for AWS and Azure.

miyunari:
As discussed with @poncovka [here ](https://github.com/redhatcloudx/transformer/pull/566#discussion_r1211891001) we decided to just keep the original google selflink, because we use it in the frontend anyway. 

closes #588


poncovka (approved):
Thank you! Is there also a pull request for the frontend? This change will probably break it.

tests/transformer/testdata/expected/google/global/rhel_7_x86_64
```diff
@@ -1 +1 @@
-{"name": "RHEL 7 X86_64", "arch": "X86_64", "version": "7", "imageId": "rhel-7-v20230306", "date": "2023-03-06T12:57:17.210-08:00", "selflink": "https://console.cloud.google.com/compute/imagesDetail/projects/rhel-cloud/global/images/rhel-7-v20230306"}
+{"name": "RHEL 7 X86_64", "arch": "X86_64", "version": "7", "imageId": "rhel-7-v20230306", "date": "2023-03-06T12:57:17.210-08:00", "selflink": "https://www.googleapis.com/compute/v1/projects/rhel-cloud/global/images/rhel-7-v20230306"}
```
FKolwa, line 1: Ah those are actually not the same thing!
Compare the output when accessing the console.cloud.google.com link (currently used in front-end) with the googleapis.com link:

https://console.cloud.google.com/compute/imagesDetail/projects/rhel-cloud/global/images/rhel-7-v20230306

https://www.googleapis.com/compute/v1/projects/rhel-cloud/global/images/rhel-7-v20230306

One of them is the marketplace link that can be used to run the image in the cloud console and the other is the API.

FKolwa: > Thank you! Is there also a pull request for the frontend? This change will probably break it.

Indeed! We need to have that change in front-end before this PR gets merged. Did you discuss how you want to proceed?

poncovka: > I've checked the front-end: We use both. The selflink is referring to the cloud console link and the api link is the reference you use for the CLI. As we need both in frontend we might think about supplying both or parse one from the other (status quo). GCP is the only cloud provider that uses this schema AWS and Azure both use distinctive IDs for the image rather than API calls.

Right now, the frontend recreates the original selflink that you replace here in the transformer, so that's why I suggested this change. See https://github.com/redhatcloudx/transformer/pull/566#discussion_r1211891001. However, yes, keeping both is also an option.

FKolwa: @poncovka Great! Yes that clarifies my questions ;) 
As long as we need to parse one of them, the direction proposed in here makes total sense. Let's talk about possible front-end implementations next.

@miyunari Could you block this PR from being merged as long as we don't have the proper front-end changes in place?

FKolwa: @miyunari @poncovka As we've discussed on Monday I've been looking into these changes and have created a ticket to summarize everything: https://github.com/redhatcloudx/transformer/issues/588

As is, the PR doesn't solve the entire issue and will only lead to more patch-work tasks.
It would also lead to confusion when working with the formatted data outside of the CID front-end context.
Instead of only  providing either the console or the api URL, the image data should contain both to stay consistent with the other cloud providers.

miyunari:
@poncovka @FKolwa please have a look again :)

poncovka (approved):
Looks good to me. Thank you! :-)

FKolwa (approved):
LGTM!

PR #582, opened 2023-06-05
add v2/all endpoint

### LINKED ISSUE #561: Hot loaded list of image details
We have decided to move away from our `image-names` list for our search functionality and instead use a list of all image details for it. So we need to add the new list to our transformer code.

blocked by #560 
blocked by #492

miyunari:
closes #561


src/cloudimagedirectory/transform/transform.py
```diff
@@ -290,4 +290,22 @@ def run(self, data):
...
+        entries = [x for x in data if not x.is_raw() and not x.is_provided_by("idx")]
+
+        results = []
+
+        for entry in entries:
+            results.append(entry.content)
+
+        results.sort(key=lambda x: x["name"], reverse=False)
+
+        return [connection.DataEntry("v2/all", results)]
```
poncovka, line 311: Oh, so it is for all images from all cloud providers. Won't it be confusing? These data structures will always look a little different (or at least contain different types of values). It also means that the image data have to be extended, because (from what I remember) they don't contain the cloud provider, the region and sometimes the architecture.

poncovka (commented):
Maybe I would focus on dropping `--image.api.version` first and added the new endpoint in another pull request. That's a lot of changes that are not really related.

src/cloudimagedirectory/transform/transform.py
miyunari: > It also means that the image data have to be extended, because (from what I remember) they don't contain the cloud provider, the region and sometimes the architecture.

That is a valid point. I started with a first PR of two, which divides this into two parts. The first will remove the version flag and add a version prefix to the reference path. The 2nd one will then extend our schema by region, cloud provider and whatever else is missing. I will double check :smiley: 
1. https://github.com/redhatcloudx/transformer/pull/585

src/cloudimagedirectory/transformer.py
```diff
@@ -80,6 +80,7 @@ def run(
             transform.TransformerIdxListImageLatestGoogle,
             transform.TransformerIdxListImageLatestAWS,
             transform.TransformerIdxListImageLatestAZURE,
+            transform.TransformerV2All,
```
miyunari, line 83: For some reason - as soon as I add this transformer - the google test starts to fail with the following message: 
```
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― test_google_transformer_command ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
[gw3] linux -- Python 3.11.2 /home/miyunari/.cache/pypoetry/virtualenvs/cloudimagedirectory-qP5mipiD-py3.11/bin/python

runner = <click.testing.CliRunner object at 0x7fbec1214e90>
tmp_path = PosixPath('/tmp/pytest-of-miyunari/pytest-6/popen-gw3/test_google_transformer_comman0')

    def test_google_transformer_command(runner, tmp_path):
        """Verify that we can transform Google data."""
        runner.invoke(
            transformer.run,
            [
                "-f",
                "tests/transformer/testdata/input/raw/google/all.json",
                "-op=.",
                f"-dp={tmp_path}",
                "--filter.until=none",
            ],
        )
    
        # Ensure the directory was made.
>       assert os.path.isdir(f"{tmp_path}/v1/google/global")
E       AssertionError: assert False
E        +  where False = <function isdir at 0x7fbede203c40>('/tmp/pytest-of-miyunari/pytest-6/popen-gw3/test_google_transformer_comman0/v1/google/global')
E        +    where <function isdir at 0x7fbede203c40> = <module 'posixpath' (frozen)>.isdir
E        +      where <module 'posixpath' (frozen)> = os.path

tests/transformer/google_test.py:22: AssertionError

 tests/transformer/google_test.py ⨯                                                                                                             50% █████     
 tests/transformer/test_all.py ✓ 
```
Any idea what could cause this? Or how to debug this further? :/

miyunari:
I added the missing `region` and `provider` - I didn't see anything else missing. The `architecture` is part of our standard schema, so it's already in the data :)

FKolwa (approved):

src/cloudimagedirectory/transform/transform.py
```diff
@@ -290,3 +291,32 @@ def run(self, data):
...
+        """Sort the raw data."""
+        # NOTE: Verify that the data is not raw.
+        entries = [x for x in data if not x.is_raw() and not x.is_provided_by("idx")]
+
+        results = []
+
+        for e in entries:
+            entry = copy.deepcopy(e)
+            filename = entry.filename.split("/")
+            if len(filename) < 3:
```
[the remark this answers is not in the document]
miyunari: Yes, that's the filename, so it's looking like this for example - entry filename: `v1/aws/ap-northeast-2/rhel_8.5_hvm_arm64_hourly2`

PR #585, opened 2023-06-06
remove version flag and prepare for multiple versions

### LINKED ISSUE #584: Remove version flag and prepare for multiple versions

miyunari:
closes #584


tests/transformer/testdata/expected/v1/idx/list/sort-by-date/0
```diff
@@ -0,0 +1 @@
+[{"name": "RHEL 7 X86_64", "date": "2023-03-06", "provider": "google", "ref": "v1/google/global/rhel_7_x86_64", "arch": "X86_64", "region": "global"}, {"name": "RHEL 6.10 hvm x86_64 Hourly2", "date": "2021-03-18", "provider": "aws", "ref": "v1/aws/af-south-1/rhel_6.10_hvm_x86_64_hourly2", "arch": "x86_64", "region": "af-south-1"}, {"name": "osa osa_311 x64", "date": "2020-01-15", "provider": "azure", "ref": "v1/azure/global/osa_osa_311_x64", "arch": "x64", "region": "global"}]
```
miyunari, line 1: This PR adds `v1/` in front of the `ref` to prepare the usage of `v2`. @FKolwa could you adapt our frontend accordingly? :)

FKolwa: Sure! Does this automatically remove everything we currently have in our S3 bucket which is not within the v1 folder structure or could this be deployed async?

FKolwa (approved):
LGTM! Thanks for handling the API Madness.

tests/transformer/testdata/expected/v1/idx/list/sort-by-date/0
miyunari: No. There is no change to the folder structure. The folder that gets synced is `/images` instead of only containing `v1` it will contain e.g. `v2` soon.

I don't know how the references to our image details are currently handled in the frontend. But I assume we use some kind of `root + ref` from the provided lists. e.g. `/images/v1/ + google/global/rhel_7_x86_64`. Previously that did point us to the correct image details. e.g. https://imagedirectory.cloud/images/v1/google/global/rhel_7_x86_64
Now ref contains the version information and we would need to remove it from the root prefix. Otherwise we end up with something like this: `images/v1/ + /v1/google/global/rhel_7_x86_64`. Does that make sense? :)

PR #597, opened 2023-06-07
Resolve shared memory issues in pytest execution

### LINKED ISSUE #596: poetry pytest fails when tests are executed on the same worker
For more details, see here:
https://github.com/redhatcloudx/transformer/pull/582#discussion_r1221451563

miyunari:
This PR addresses shared memory issues encountered when running `poetry run pytest`. By introducing `pytest-xdist` and increasing the number of workers to 16 (which is higher than the tests we currently execute), we are able to mitigate the problem and improve test execution reliability. Since 16 instead of 2-6 workers are created, the execution time increases slightly.

**Changes Made**
- Use `runner.Invoke` in all e2e tests.
- Added `pytest-xdist` as dependency to enable parallel test execution.
- Configure `pyproject.toml` to set the number of workers to 16 using `-n 16` arg.
- Compare transformer exit code and provide more debug details if its not `0`.

Closes #596


FKolwa:
Uh wow, is this issue once again caused by tmp_path? If so, is it worth keeping it?

miyunari:
Originally I had the same thought, but looking into `/tmp`, as @poncovka recommended showed that every testcase runs in its own folder and everything is fine - so it was not the fault of the `tmp` folder.
The point here is, that we call [transfomer.run](https://github.com/redhatcloudx/transformer/blob/17fa0b566eaaf701acc52a39da5ed6cb88c06ca3/tests/transformer/aws_test.py#LL11C17-L11C17) here and in other tests. Inside of the run method a [pipeline](https://github.com/redhatcloudx/transformer/blob/17fa0b566eaaf701acc52a39da5ed6cb88c06ca3/src/cloudimagedirectory/transformer.py#L69) is created. The pipeline is for e.g. initialized with [this 3 transformers](https://github.com/redhatcloudx/transformer/blob/17fa0b566eaaf701acc52a39da5ed6cb88c06ca3/src/cloudimagedirectory/transformer.py#L71) that run first. 
https://github.com/redhatcloudx/transformer/blob/17fa0b566eaaf701acc52a39da5ed6cb88c06ca3/src/cloudimagedirectory/transform/transform.py#L31-L32
Here we see that all passed transformers are created and added to the internal list of transformers.

Since its in-line configured this list should always have then length of `3` right now. And that is the case if we execute the transformer once on a worker. What I did was adding a `print(len(self.transformers))` and I could observe how the number of transformers started to raise after every 2-3 test execution. It seemed to be a bit flaky. Sometimes the last failing test had 3-4 times the containers registered that are marked here.
https://github.com/redhatcloudx/transformer/blob/17fa0b566eaaf701acc52a39da5ed6cb88c06ca3/src/cloudimagedirectory/transformer.py#L71-L75

That's why I added multiple workers, so every test can run isolated on it's own worker node. And now they don't interfere anymore or cause other trouble.

FKolwa:
@miyunari That's interesting! So do I get this right: 
- The transformer cli runner should always create these 3 transformer units (AWS, Azure, Google)
- During the run this array might contain more or less of these transformer units
- The test executions somehow impact the transformer 

Does this mean that the transformer runner that is used within the pytest testcases is reused across the entire test execution?

FKolwa:
Thanks for the insight, this was really well explained! I'm so used to have tests run in encapsulated environments that test cases that share state beyond their own scope boggles my mind. I really want you to be unblocked on this and as far as I understand, this is only happening during testing so let's get this merged. One more question regarding pytest though: Could it potentially happen that even though we have more than enough worker nodes two or more test instances will share one node?

FKolwa (approved):

miyunari:
> ... as far as I understand, this is only happening during testing ...

Right, because in a normal run the pipeline is initialised once. So there is no way that another pipeline instance could cause side effects. :) 

> Could it potentially happen that even though we have more than enough worker nodes two or more test instances will share one node?

No, I do not think so. But it would become an issue as soon as we had more than 16 tests creating a pipeline.

---

**Update**


We can keep the parallel execution, but theoretically it would work without this kind of isolation since the last fix (https://github.com/redhatcloudx/transformer/pull/597/commits/2023ee8f25ceaa12b336c542127288b22c09f543). @KnutZuidema pointed out that in Python, if you define a variable inside a class but outside a method or constructor, it becomes a class variable or static variable. This means that the variable is shared by all instances of the class. This caused the side effect we observed :)

FKolwa:
Sweet! I very much prefer this solution. Thanks for changing this.

src/cloudimagedirectory/transform/transform.py
```diff
@@ -14,10 +14,6 @@
 class Pipeline:
     """Builds a pipeline of transformer tasks."""
 
-    transformers: list[Callable] = []
-    filter_funcs: list[Callable] = []
-    idx_generators: list[Callable] = []
```
poncovka, line 19: Have you checked other classes to make sure that they don't define class attributes as well?

tests/transformer/aws_test.py
```diff
@@ -18,6 +18,13 @@ def test_aws_transformer_command(runner, tmp_path):
         ],
     )
 
+    if "0" != f"{result.exit_code}":
```
poncovka, line 21: You should be able to do just `assert result.exit_code == 0`. Pytest should print the exit code (and possibly the result) for you if the check fails. Ideally, you could write a helper function that would run the transformer and do some default checks of the result. You can see in this commit that there is a pattern that you repeat in each test. That's problematic when you want to do some systematic changes in the tests (like this one).

tests/transformer/idx_images_by_name_test.py
```diff
@@ -1,20 +1,36 @@
...
-        "${PWD}/tests/transformer/testdata/input/raw/azure/eastus.json,"
-        "${PWD}/tests/transformer/testdata/input/raw/google/all.json,"
-        "${PWD}/tests/transformer/testdata/input/raw/aws/af-south-1.json"
-        " -op=${PWD} -dp=${PWD}/tests/transformer/testdata"
-        " --filter.until=none"
+    result = runner.invoke(
+        transformer.run,
+        [
+            "-f",
+            "tests/transformer/testdata/input/raw/google/all.json,tests/transformer/testdata/input/raw/aws/af-south-1.json,tests/transformer/testdata/input/raw/azure/eastus.json",
```
poncovka, line 15: I don't know if your linter will allow it, but you should be able to split the line to multiple strings like this:

```
        [
            "-f",
            "tests/transformer/testdata/input/raw/google/all.json,"
            "tests/transformer/testdata/input/raw/aws/af-south-1.json,"
            "tests/transformer/testdata/input/raw/azure/eastus.json",
            ...
    ]
```

The trick is that Python automatically concatenates string literals into one string. For example:

```
>>> x = "aaa" "bbb" "ccc"
>>> print(x)
aaabbbccc
```

poncovka (approved):
I have some suggestions, but otherwise great job catching this bug :-)

tests/transformer/aws_test.py
miyunari: Maybe I don't fully got it, but this thing is there to print `result.output` in a case of an error. This output helped me to understand the reason for this non zero error code. Does this make sense? :)

miyunari:
@poncovka I know that this is still open https://github.com/redhatcloudx/transformer/pull/597#discussion_r1227767780, I merge the PR now anyway to be no longer blocked for the `/all` endpoint. If we want to change something after that, I'll be happy to adjust it again :)

tests/transformer/aws_test.py
poncovka: That totally makes sense, but it can be simplified :-) For example, you can achieve the same thing with the `assert condition, message` statement:

```python
assert result.exit_code == 0, f"expected no error, but got code {result.exit_code} and output:\n{result.output}"
```

Why it is better?
* Don't convert values to other types if it is not necessary. If you expect an integer, let's check an integer.
* The body of the `if` statement is not run when the test succeeds. However, the test should usually succeed. That means that this code doesn't have to work and you won't know that until the test breaks. Basically, you have no code coverage.
* It is easier to read.

miyunari: Ah nice! Thanks a lot - I will address this :smile: I didn't know it was possible that way :+1:

miyunari: Follow up PR: https://github.com/redhatcloudx/transformer/pull/612

PR #615, opened 2023-06-21
add API v2 list os endpoint

### LINKED ISSUE #589: API v2: list all available operating systems
Create a list with all available operating systems for the images in it. 
`/v2/os/list`

miyunari:
In this PR a new endpoint `/v2/os/list` was added. 
At the moment I decided not to focus the different os names and just go with `rhel` 
(I am not sure if they are really all the same). 
Should we leave it that way or would it be better to have all the different names listed as different operating systems?

Here are the different os names:
```
rh-ocp-worker
rh-oke-worker
rh-opp-worker
rh-rhel
rhel-arm64
rhel-byos
rhel-raw
rhel-sap-apps
rhel-sap-ha
rh
```

closes #589


FKolwa:
Okay the list is a bit of a mixed bag. I can see a few different "products" in your list (SAP, SAP-Apps) as well as "build targets" (rhel-arm64, rhel-raw), an "offer" (rhel-byos) and whatever stuff like rh or rh-rhel are. I assume that these come from brew package names that are used in the release process. This will be a bit tricky to filter but I'd really like to see an endpoint we can use to just retrieve individual products like SAP and SAP-Apps images (or other products like JBoss EAP etc.). So I'd vote for having these items split into their own lists to make it as easy as possible to just retrieve individual products. I'm also unsure if "OS" is the best name for this category.

miyunari:
Hmm as far as I understood `rhel-sap-apps` and `rhel-sap-ha` are the names of different operating systems:

https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent?hl=de
grafik

We extract the `rh-rhel`, `rh-ocp-worker`, `rh-oke-worker`, `rh-opp-worker`, `rhel-arm64`, `rh` and `rhel-raw`  from the `SKU`'s in Azure and add it to the `name` that's why they are here in the list. The `rhel-byos`, `rh-rhel` is out of the `offers`s in Azure, which we also add to the name in the `format-azure` file. So the ones out of the `offers` can probably get reduced to `rhel`, but I guess the ones extracted from the `SKU`s are also different operating systems, right? 

https://github.com/redhatcloudx/transformer/blob/7cb972c3e9b118e9247f5168f261202af2683250/src/cloudimagedirectory/format/format_azure.py#L68

And to the other point with the `product` endpoint. I think, when we want to include other `OS`es in the future like `fedora` then we should keep it to operating systems and not divide into different products. Since my understanding is that `fedora-core-os`, `fedora-iot` and `fedora-workstation` are different operating systems. What do you think?

FKolwa:
Ah now I see where those are coming from. 
Yeah I think the Google cloud page doesn't do a good job when it comes to covering our cloud products, or let's say they use a different language when they talk about cloud images. The OS is in all of those images the same (RHEL) which confused me a little. What Google calls OS is what AWS and Azure calls offers. And yes I'm not suggesting to replace OS with "product" I'd rather suggest that it would make sense to split this into two different levels e.g. "os/(OS)/offer/(offer)/..." So for example "os/RHEL/offer/rhel-sap/..." What do you think? This would require an additional level in our API hirarchy but it would (at least to me) be a little more straightforward. In the future this could of course also be extended to os/fedora/

miyunari:
Ah ok - yes, I think that would make sense! :smiley: Then let's just do that and I will extend it in the documentation and create a new issue for that new product endpoint :+1: 

But to be sure - for this PR it means that we let it as it is and everything we have is listed as `rhel`, right?

FKolwa:
Great, yes I think we can leave the PR as is and proceed in a new issue. Thanks for getting this done!

miyunari:
Here is the new issue for the discussed endpoint https://github.com/redhatcloudx/transformer/issues/620

src/cloudimagedirectory/transform/transform.py
```diff
@@ -320,3 +320,86 @@ def run(self, data):
...
+                os_list["rhel"] += val
+                os_list.pop(os)
+            elif os == "rh":
+                os_list["rhel"] += val
+                os_list.pop(os)
+
+        for os, val in os_list.items():
+            desc = "no description"
+            if os in self.description:
+                desc = self.description[os]
```
poncovka, line 391: You can use [dict.get](https://docs.python.org/3/library/stdtypes.html#dict.get) to handle the default value:

```
desc = self.description.get(os, "no description")
```

poncovka (commented):
I have some suggestions, but they can be addressed later :-)

miyunari:
I've tried to address all the suggestions - could you have a look again @poncovka and @FKolwa ? :smile:

poncovka (commented):
It looks much better. Thanks! :-)

FKolwa (approved):
LGTM!

PR #631, opened 2023-06-29
transformer: Filter images by their unique references instead of names

poncovka:
If there are multiple images with the same reference, use only the latest image. For example, the `v1/aws/us-east-1/rhel_9.1.0_hvm_x86_64_hourly2` reference can be used by only one image provided by the CID and it should be the latest one.

### DIFF
--- a/src/cloudimagedirectory/filter/filter.py
+++ b/src/cloudimagedirectory/filter/filter.py
@@ -32,16 +32,16 @@ def FilterImageByLatestUpdate(latestDate: pd.Timestamp) -> Callable:
     ]
 
 
-def FilterImageByUniqueName() -> Callable:
-    """Filter latest images with unique names."""
-    print("filter images by unique names")
-    return _filter_by_unique_names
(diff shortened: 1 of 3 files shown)


miyunari (commented):
~test~

src/cloudimagedirectory/filter/filter.py
```diff
@@ -50,22 +50,22 @@ def _filter_by_unique_names(data):
         if entry.content is None:
             continue
 
-        # Compare the data entry with the last inserted entry of
-        # the same name. If the new entry is older, do nothing.
-        name = entry.content["name"]
+        # Compare the data entry with the last inserted entry with
+        # the same reference. If the new entry is older, do nothing.
+        ref = entry.filename
```
miyunari, line 55: I cant follow 100%, could you help me and explain the details a bit? :)

My understanding is that the filename is always unique, so that we never need to compare it with other image data and never filter something? :see_no_evil:

PR #632, opened 2023-06-29
Update the poetry.lock file

poncovka:
Poetry introduced a new format of the `poetry.lock` file in [1.3.0](https://github.com/python-poetry/poetry/releases/tag/1.3.0). Unfortunately, there is no support that would allow to enforce a minimal version of poetry or the lock file. It means that these versions have to be checked during reviews (to make sure that the format won't be flipped back to the old one).

Also, I used the [deptry](https://github.com/fpgmaas/deptry) tool to identify unused dependencies and removed them from the `pyproject.toml` and `poetry.lock` files. That might help a bit with the flood of dependecy updates that you have for this repository.


major:
Fixed this in #662!

PR #666, opened 2023-07-18
Enable mypy

major:
Some of the "fixes" applied in the second commit still need more work but this should get us going with mypy a little faster.


major:
@miyunari PR #666 is spooky. 👻

miyunari (approved):
LGTM! Thanks for doing this :)

PR #667, opened 2023-07-18
Dataentry updates

major:
Simplify the is_raw method into a property to avoid using a callable.

Also adjust `test_get_content` to store the parsed JSON in a different
variable to avoid reusing `content`. That confused mypy a little.

### DIFF
--- a/src/cloudimagedirectory/connection/connection.py
+++ b/src/cloudimagedirectory/connection/connection.py
@@ -1,9 +1,11 @@
 """Manage content in the S3 bucket."""
+from __future__ import annotations
+
 import json
 import os
 import pathlib
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Any
 
 
 class OriginPathDoesNotExist(Exception):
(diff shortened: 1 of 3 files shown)


src/cloudimagedirectory/connection/connection.py
```diff
@@ -2,24 +2,23 @@
...
 class DataEntry:
     """Handles a file from the bucket."""
 
-    # TODO: Set Any to a specific type.
-    def __init__(self, filename: str, content: Any) -> None:
-        """Constructor for DataEntry class."""
-        self.filename = filename
-        self.content = content
+    filename: str
+    content: str
```
miyunari, line 19: How does this work? Isn't this field used in other places as a dictionary? :D

major: @miyunari so in a dataclass, you don't need an `__init__` for setting class arguments. You just specify them at the top of the class definition and it takes them in if they're provided.

miyunari: Oh, I meant `content` was a dictionary before, so of type `any` and now you changed it to `string`. Doesn't that cause trouble in other places, where I used `content` as a dict before? :)

major: So the issue was that `test_get_content` was re-using `content` which was typed as a string. That confused `mypy`. I fixed it and switched the `content` argument to be a dictionary or `None`.

major:
@miyunari Test coverage went down 0.15% in the patch but improved overall. 🙃 That's where the red :x: is coming from.

PR #668, opened 2023-07-18
Remove Connection class

major:
The class was only inherited by ConnectionFS and only used in two tests.


major:
I'll go ahead and merge this one since it doesn't really do anything to our existing code.·

PR #676, opened 2023-07-20
format_google: Add test coverage

major:
This brings us to 💯 test coverage for `format_google`.

Turn the parser into a simple dataclass to avoid regular expressions and worrying with amd64/sap bumping into each other. Also split out some of the selflink generation code and avoid re-use of variables.


major:
Oh boy, this takes us over 90% coverage! 👏👏👏

miyunari (approved):

Issue #671, opened 2023-07-19
API v2: Create basic structure


miyunari:
It would be great to create the basic structure for the second API version, before writing all the endpoints. 
This includes implementing the `AWS`, `Azure` and `Google` Transformer first.

- [x] TransformerAzureV2RHEL
- [x] TransformerAWSV2RHEL
- [x] TransformerGoogleV2RHEL

PR #682, opened 2023-07-24
V2 create basic structure #671

miyunari:


src/cloudimagedirectory/transform/transform.py
```diff
@@ -295,27 +295,144 @@ def run(self, data: type[Transformer]) -> list:
...
+                if "rhel" in content["name"]:
+
+                    image_data = format_google.image_rhel(content)
+                    image_name = image_data["name"].replace(" ", "_").lower()
+                    region = "global"
+                    os_name = "rhel"
+                    provider = "google"
+                    version = image_data["version"]
+                    # NOTE: Due to consistency issues between the cloud providers and the fact
+                    # that they do not all have unique numbers to identify their images, we decided
```
FKolwa, line 401: Are those names that you're using for hashing unique? 
Just asking cause the created hash will be deterministic. 
Not that we're running into the same id overlap with those hashes.

miyunari: Not in general unique, but in their specific path, they are.

Before we had the identification based on the image name. Now it's based on the full path. 
For example if we have two equal image names `XXXXXXXXXXXXXXXXXXXXXX` and those are different images, it should have no conflict because they are stored in different paths :) 

like this:
```bash
/v2/os/rhel/provider/azure/version/7-8/region/eu-west-1/image/XXXXXXXXXXXXXXXXXXXXXX
/v2/os/fedora/provider/aws/version/9/region/us-east-22/image/XXXXXXXXXXXXXXXXXXXXXX
````

File: src/cloudimagedirectory/transform/transform.py [hunk 2]
```diff
@@ -295,27 +294,143 @@ def run(self, data: type[Transformer]) -> list:
...
             entry = copy.deepcopy(e)
+
             filename = entry.filename.split("/")
             if len(filename) < 3:
                 print("warn: could not determine region or provider of image: " + entry.filename)
                 continue
 
-            entry.content["provider"] = filename[1]
-            entry.content["region"] = filename[2]
+            entry.content["provider"] = filename[4]
```
FKolwa, line 432: This is totally fine but I can imagine that these array access values might be confusing in the long run. Maybe we can use namedtubles in the future? It would at least make accessing the individual fields easier.

Just an example:

```
  from collections import namedtuple
  
  APIPath = namedtuple('APIPath', ['api_version', 'distro', 'provider', 'version','region','image_id'])
  
  split_path = "v2/rhel/google/9.0.0/us-east-1/asdasdasdadaf".split('/')
  api_path = APIPath(*split_path)
  
  print(api_path.api_version)
  print(api_path.image_id)
```
That api_path could be a property of the DataEntry object or something.

src/cloudimagedirectory/transform/transform.py
FKolwa: Okay cool! I guess we're not running into any edge cases in which we need to lookup images by their imageId value alone, right?

miyunari: Not so far. In that case we would need another endpoint to lookup images by ID. For example `/v2/image-ids/list`, right?
That would return a list of image IDs and their full paths.

src/cloudimagedirectory/transform/transform.py
miyunari: I agree, that looks way better to maintain. :smile: Is it ok, when I adapt in another pr? :)

miyunari: I created a new issue for it :) https://github.com/redhatcloudx/transformer/issues/685

FKolwa: Absolutely, thanks for taking care of this.

src/cloudimagedirectory/transform/transform.py
FKolwa: Yeah, as long as we avoid that we should be fine.

FKolwa (approved):

Issue #424, opened 2023-03-23
transformer: add Unit tests


miyunari:
TODO

- [ ]  Pipeline
- [ ] AWS Transformer
- [ ] Azure Transformer
- [ ] Google Transformer

PR #686, opened 2023-07-25
V2 list provider by OS #590

### LINKED ISSUE #590: API v2: list all available providers for given os
Create a list with all available providers for the given product.

`/v2/os/<os>/product/<product>/provider`

Example:
`/v2/os/rhel/product/rhel-sap-apps/provider`

miyunari:
closes #590

### DIFF
--- a/src/cloudimagedirectory/transform/transform.py
+++ b/src/cloudimagedirectory/transform/transform.py
@@ -489,3 +489,37 @@ class TransformerV2ListOS(Transformer):
 
         # NOTE: Add /list suffix to prevent collision with "os" folder.
         return [connection.DataEntry("v2/os/list", results)]
+
+
+class TransformerV2ListProviderByOS(Transformer):
+    """Generate a list for all available providers of a specific os."""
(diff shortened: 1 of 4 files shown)


major (approved):
I'd like to see us find a way to interrogate these transformers with more specific tests rather than examining the full entire output, but this looks good enough to go for now. I can help with test coverage and mypy issues.

miyunari:
> I'd like to see us find a way to interrogate these transformers with more specific tests rather than examining the full entire output, but this looks good enough to go for now. I can help with test coverage and mypy issues.

Yes, you are absolutely right with this, I think we all see it that way. Therefore I created an [issue](https://github.com/redhatcloudx/transformer/issues/424) a while ago, but currently we needed to prioritize and decided to add them later and focus on the second API version this quarter :/ At the moment the code changes so fast, that we would need to adjust them multiple times, that was another reason, we thought it would be better to wait until the big changes are made :/  But we should definitely add those in the long run!

PR #688, opened 2023-07-26
V2 list region by version #593

### LINKED ISSUE #593: API v2: list all regions selected
Create a list with all regions which are selected.

`/v2/os/<os>/product/<product>/provider/<provider>/version/<version>/region`

Example:
`/v2/os/rhel/product/rhel-sap-apps/provider/aws/version/9-1/region`

miyunari:
closes #593


src/cloudimagedirectory/transform/transform.py
```diff
@@ -564,3 +564,50 @@ def run(self, data: type[Transformer]) -> list:
                     connection.DataEntry(f"v2/os/{os}/provider/{provider}/version/list", version_map[provider])
                 )
         return results
+
+
+class TransformerV2ListRegionByVersion(Transformer):
+    """Generate a list for all available regions for one version."""
+
+    def run(self, data):
```
major, line 572: Add `# noqa: C901` here to silence the complexity linter. It would be good to revisit how we could make this less complex later. :wink:

major (changes requested):
Just one minor linting issue.

src/cloudimagedirectory/transform/transform.py
miyunari: Ah right, I added it :)

major (approved):
Thanks for catching the `no_type_check` in there. This looks okay now but we will need to work on this complexity later. 😉

miyunari:
Yes, alright - I created an issue, so we don't forget about it :) https://github.com/redhatcloudx/transformer/issues/689

PR #691, opened 2023-07-26
Convert DataEntry to a dataclass

major:
This simplifies the DataEntry class a bit and allows us to have a really handy `__repr__` output whenever we need to print out the contents of the class.

Also, this fixes a mypy issue where `content` is recast from a string to a dictionary when `json.loads()` reads it.


miyunari (approved):
LGTM! :)

PR #692, opened 2023-07-26
transform: Fix type checks

major:
Remove all of the `@no_type_check` markers and use the `connection.DataEntry` type with all transformers. There are still some issues with passing a list to DataEntry when it expects to have a dictionary or `None` as its second argument.

Because if mypy isn't happy, nobody is happy. 😉

### DIFF
--- a/src/cloudimagedirectory/transform/transform.py
+++ b/src/cloudimagedirectory/transform/transform.py
@@ -82,11 +82,9 @@ class TransformerIdxListImageLatest(Transformer):
 
     # TODO: The ruff linter complains that this method is too complex. We might be able
     # to break it up into more manageable chunks.
-    # TODO: Mypy says that 'data' below is not iterable. This needs to be fixed later.
-    @no_type_check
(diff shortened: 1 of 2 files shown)


src/cloudimagedirectory/transform/transform.py
```diff
@@ -291,13 +280,15 @@ def run(self, data: type[Transformer]) -> list:
 
         results.sort()
 
-        return [connection.DataEntry("v1/idx/list/image-names", results)]
+        # TODO: DataEntry expects a dictionary as its second argument, but this is
+        # passing a list. Should we update DataEntry to accept a list?
```
miyunari, line 284: Actually it should accept everything that can be serialized to JSON. Should it then be of type `any` or is there a type in Python like `serializable`? :)

major: Ah, I was able to adjust it to be `list | dict | None`, which isn't great, but it gets us over this problem for now.

miyunari (approved):

PR #694, opened 2023-07-26
V2ListRegionByVersion: Try simplifying

major:
Remove some loops and build out a DataEntry-like structure from the start. Then convert it to data entries.

The complexity came down enough that we can remove the ruff ignore line.


miyunari (approved):
Wow! Thanks a lot! This looks way better :heart_on_fire: :open_mouth:

PR #727, opened 2023-08-28
Instrument the transformer using the opentelemetry-sdk

### LINKED ISSUE #728: Instrumentation of OpenTelemetry
We should instrument our transformer data with the openTelemetry sdk to get a better overview of what's happening inside our project and to detect unwanted behaviour earlier.

miyunari:
This PR instruments:
- the v2 provider transformer
- the v2 list all endpoint transformer
- all Image filters 
 
While I have still an open question how to call the counter add method when data gets filtered using lambda.

As a next step, we have to add a otel exporter to store and visualize the data.

closes #728


src/cloudimagedirectory/filter/filter.py
```diff
@@ -23,6 +36,13 @@ def FilterImageByLatestUpdate(latestDate: pd.Timestamp) -> Callable:  # type: ig
     print(f"filter images by latest date: {latestDate}")
     latestDate = latestDate.replace(tzinfo=pytz.UTC)
 
+    filtered_image_by_latest_update_counter = meter.create_counter(
+        name="filtered_image_by_latest_update_counter",
+        description="Counts the number of filtered images by latest update",
+        unit="1",
+    )
+    # TODO: Call method every time the lambda skips an element.
+    # filtered_image_by_latest_update_counter.add(1, {"latest_date", latestDate})
```
miyunari, line 45: Whats the best way to call the counter add method every time an item is filtered?
```python
return lambda data: [d for d in data if d.content is not None and get_utc_datetime(d.content["date"]) > latestDate]
```

src/cloudimagedirectory/filter/filter.py
```diff
@@ -15,6 +19,14 @@ def get_utc_datetime(date_string: str) -> pd.Timestamp:  # type: ignore[no-any-u
...
     print("filter images by filename: " + word)
+
+    meter.create_counter(
+        name="filtered_image_by_filename_counter",
+        description="Counts the number of filtered images by filename",
+        unit="1",
+    )
+    # TODO: Call method every time the lambda skips an element.
+    # filtered_image_by_filename_counter.add(1, {"word", word})
     return lambda data: [d for d in data if not d.filename.lower().__contains__(word.lower())]
```
poncovka, line 30: I am afraid you will have to replace the lambda functions and get rid of the list generators. I don't think there is a nice trick that would allow you to do what you need. You can create an inner function instead and return that one.

```
def f(data):
    result = []
    for d in data:
        ...
    return result

return f
```

miyunari: @poncovka , I tried to replace the lambda functions with inner functions. 
Could you have a look again, if it makes sense how I did it and if it is in a "python way" :sweat_smile: ?

src/cloudimagedirectory/filter/filter.py
```diff
@@ -15,15 +19,44 @@ def get_utc_datetime(date_string: str) -> pd.Timestamp:  # type: ignore[no-any-u
...
     print("filter images by filename: " + word)
-    return lambda data: [d for d in data if not d.filename.lower().__contains__(word.lower())]
+
+    filtered_image_by_filename_counter = meter.create_counter(
+        name="filtered_image_by_filename_counter",
+        description="Counts the number of filtered images by filename",
+        unit="1",
+    )
+
+    def _filter_image_by_filename(data: str) -> list:
```
poncovka, line 29: I think `data` is a list of entries, isn't it?

src/cloudimagedirectory/filter/filter.py
```diff
@@ -15,15 +19,44 @@ def get_utc_datetime(date_string: str) -> pd.Timestamp:  # type: ignore[no-any-u
...
+        result = []
+        for d in data:
+            if not d.filename.lower().__contains__(word.lower()):
```
poncovka, line 32: It is better to use `in` instead of `__contains__`. For example: `if word.lower() not in d.filename.lower()`

src/cloudimagedirectory/filter/filter.py
```diff
@@ -15,15 +19,44 @@ def get_utc_datetime(date_string: str) -> pd.Timestamp:  # type: ignore[no-any-u
...
+
+    def _filter_image_by_latest_update(data: str) -> list:
+        result = []
+        for d in data:
+            if d.content is not None and get_utc_datetime(d.content["date"]) > latestDate:
+                result.append(d)
+            else:
+                filtered_image_by_latest_update_counter.add(1, {"latest_date", latestDate})
+        return result
+    return _filter_image_by_latest_update
```
poncovka, line 59: Just for info, another useful way of filtering items is to use the [filter](https://docs.python.org/3/library/functions.html#filter) function. For example:

```python

def _is_latest(d) -> bool:
    return d.content is not None and get_utc_datetime(d.content["date"]) > latestDate

return lambda data: list(filter(_is_latest, data))
```

Then you can extend the `_is_latest` function in a similar way as above.

src/cloudimagedirectory/filter/filter.py
miyunari: Yes, I remember our meeting around that topic. As we used these method names in many places of the project, I would prefer to keep it here consistent with the other places and change them in an extra PR - What do you think? 
I opened an issue for that topic https://github.com/redhatcloudx/transformer/issues/733

poncovka: Yeah, in this case, it is just a bad practice, but it is definitely good to raise awareness of this issue :-)

src/cloudimagedirectory/filter/filter.py
```diff
@@ -15,15 +20,46 @@ def get_utc_datetime(date_string: str) -> pd.Timestamp:  # type: ignore[no-any-u
 def FilterImageByFilename(word: str) -> Callable:
     """Filter images by filename."""
     print("filter images by filename: " + word)
-    return lambda data: [d for d in data if not d.filename.lower().__contains__(word.lower())]
+
+    filtered_image_by_filename_counter = meter.create_counter(
+        name="filtered_image_by_filename_counter",
```
frzifus, line 25:
```suggestion
        name="filter.image.by_filename.count",
```

src/cloudimagedirectory/filter/filter.py
```diff
@@ -15,15 +20,46 @@ def get_utc_datetime(date_string: str) -> pd.Timestamp:  # type: ignore[no-any-u
...
 
 
 def FilterImageByLatestUpdate(latestDate: pd.Timestamp) -> Callable:  # type: ignore[no-any-unimported]
     """Filter images by latest date."""
     print(f"filter images by latest date: {latestDate}")
     latestDate = latestDate.replace(tzinfo=pytz.UTC)
 
-    return lambda data: [d for d in data if d.content is not None and get_utc_datetime(d.content["date"]) > latestDate]
+    filtered_image_by_latest_update_counter = meter.create_counter(
+        name="filtered_image_by_latest_update_counter",
```
frzifus, line 48:
```suggestion
        name="filter.image.by_latest_update.count",
```

src/cloudimagedirectory/filter/filter.py
```diff
@@ -39,6 +75,12 @@ def _filter_by_unique_names(data: list) -> list:
     # insertion order of the data entries.
     unique_data: dict[Any, Any] = {}
 
+    filtered_image_by_unique_names_counter = meter.create_counter(
+        name="filtered_image_by_unique_names_counter",
```
frzifus, line 79:
```suggestion
        name="filter.image.by_unique_name.counter",
```

src/cloudimagedirectory/transform/transform.py
```diff
@@ -282,6 +286,13 @@ def run(self, data: list[DataEntry]) -> list:
         return [DataEntry("v1/idx/list/image-names", results)]
 
 
+generated_image_provider_metadata_counter = meter.create_counter(
+    name="generated_image_provider_metadata_count",
```
frzifus, line 290:
```suggestion
    name="generator.image.metadata.provider.count",
```

src/cloudimagedirectory/transform/transform.py
```diff
@@ -395,9 +408,17 @@ def run(self, data: list[DataEntry]) -> list:
                     data_entry = DataEntry(path, image_data)
 
                     results.append(data_entry)
+                    generated_image_provider_metadata_counter.add(1, {"provider": provider})
         return results
 
 
+generated_image_endpoint_metadata_counter = meter.create_counter(
+    name="generated_image_metadata_count",
```
frzifus, line 416:
```suggestion
    name="generator.image.metadata.count",
```

frzifus (commented):
I'm no expert at naming things, but do the suggestions make sense? Otherwise LGTM!

Here is a link to the convention: [otel-metrics-semantic-conventions](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/)

src/cloudimagedirectory/transform/transform.py
```diff
@@ -518,7 +540,7 @@ class TransformerV2ListVersionByProvider(TransformerV2):
     """Generate a list for all available versions for a specific provider."""
 
     def run(self, data: list[DataEntry]) -> list:
-        # Start each version at a count of 0 so we can increment the counter as
```
FKolwa, line 521: I guess this was removed by mistake?

src/cloudimagedirectory/transformer.py
```diff
@@ -41,6 +41,8 @@
     help="List of predefined files to process",
 )
 def run(origin_path: str, destination_path: str, arg_files: str, filter_until: str) -> None:
+    # metrics.get_meter_provider().start_pipeline(ConsoleMetricExporter(), interval=5)
```
FKolwa, line 44: Is this a placeholder for future implementations?

FKolwa (approved):
This is a great foundation, I can't wait to see the rest of the implementation! I only got some really minor questions that have no impact on functionality. Feel free to merge once you're ready.

src/cloudimagedirectory/transform/transform.py
miyunari: lol - yes you are right! :D

src/cloudimagedirectory/transform/transform.py
```diff
@@ -518,7 +540,7 @@ class TransformerV2ListVersionByProvider(TransformerV2):
...
+        # Start each version at a count of 0 so we can increment the as
```
miyunari, line 543:
````suggestion
        # Start each version at a count of 0 so we can increment the counter as
```

src/cloudimagedirectory/transformer.py
miyunari: yes, the next step is to export the data.

src/cloudimagedirectory/transform/transform.py
```diff
@@ -540,7 +540,8 @@ class TransformerV2ListVersionByProvider(TransformerV2):
     """Generate a list for all available versions for a specific provider."""
 
     def run(self, data: list[DataEntry]) -> list:
-        # Start each version at a count of 0 so we can increment the as
+        # Start each version at a count of 0 so we can increment the counter as
+```
```
miyunari, line 544:
````suggestion
````

poncovka (commented):
Late, but it looks good to me :-)

